PR #49, opened 2017-12-18
updating license links

MLNoon:
Note: Datasets provided by the US Federal Government under public domain license unless otherwise noted on their website for datasets. All datasets here are marked as unrestricted use and access

### DIFF
--- a/LDMP/help/source/about/about.rst
+++ b/LDMP/help/source/about/about.rst
@@ -78,109 +78,108 @@ separate terms as indicated in their respective metadata.
 NDVI
 ~~~~
 
-+------------------+-----------+---------+--------+-----------+
-| Sensor/Dataset   | Temporal  | Spatial | Extent | License   |
-+==================+===========+=========+========+===========+
-| `AVHRR/GIMMS`_   | 1982-2015 | 8 km    | Global |  `NEX`_   |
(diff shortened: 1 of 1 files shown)


LDMP/help/source/about/about.rst
```diff
@@ -78,121 +78,120 @@ separate terms as indicated in their respective metadata.
...
-| Sensor/Dataset   | Temporal  | Spatial | Extent | License   |
-+==================+===========+=========+========+===========+
-| `AVHRR/GIMMS`_   | 1982-2015 | 8 km    | Global |  `NEX`_   |
-+------------------+-----------+---------+--------+-----------+
-| `MOD13Q1-coll6`_ | 2001-2016 | 250 m   | Global | `LPDAAC`_ |
-+------------------+-----------+---------+--------+-----------+
++------------------+-----------+---------+--------+-----------------+
+| Sensor/Dataset   | Temporal  | Spatial | Extent |     License     |
++==================+===========+=========+========+=================+
+| `AVHRR/GIMMS`_   | 1982-2015 | 8 km    | Global |`Public Domain`_ |
```
azvoleff: There has to be a space between the pipe and the "`" symbol. So "|`Public Domain`_ |" has to be "| `Public Domain`_ |". Can you fix this and the following links that are missing the space?

PR #99, opened 2018-03-13
Update to site MGR

mgroglich:


azvoleff:
Can you double check this? It looks like you might have moved a lot of files but not committed them. so if I merge this it will delete most of the files from the help section. See this folder: https://github.com/mgroglich/trends.earth/tree/master/LDMP/help/source to see what is in your repository. All that is left are the video and about sections.

PR #127, opened 2018-04-06
Updated tutorials

azvoleff:


azvoleff:
I'm merging this, but sorry, I need to rename the .rst files (except for the new ones) back to how they were originally named (otherwise I'll need to relink all of them again with transifex). I think it is probably best to avoid using numbers in the filenames anyways in case we move things in the future.

PR #136, opened 2018-04-06
added rst file and changes some names... will it work?

mgroglich:


azvoleff:
Almost right. But the filenames ending in .tex in pavement.py and conf.py need to match the filenames in the links in the rst files (the ones ending in .pdf). They have to be exactly the same except for the extensions. Can you fix that and then push again?

PR #226, opened 2019-11-21
 Fixing non polygon error

azvoleff:


azvoleff:
@gdaldegan - it looks like a couple of other places in data_io.py have the same "fieldNameIndex" issue that you fixed on line 286. (see lines 985, 1054, and 1126). Can you fix those cases as well, and then push a new commit with those changes also included?

gdaldegan:
sure!
looking at ir right now

PR #319, opened 2020-11-12
Adds PT and ES translations to LPKS infotext json

gdaldegan:


LDMP/calculate_urban.py
```diff
@@ -156,7 +156,7 @@ def btn_calculate(self):
         # Limit area that can be processed
         aoi_area = self.aoi.get_area() / (1000 * 1000)
         log(u'AOI area is: {:n}'.format(aoi_area))
-        if aoi_area > 25000:
+        if aoi_area > 10000:
```
azvoleff, line 159: Can you fix this line? We want to keep the aoi_area limit at 25000 I think, right? If we changed it then it wouldn't align with the message on the line below.

azvoleff (changes requested):
See comment on line 159 of calculate_urban.py

PR #347, opened 2021-01-25
Aligned task help to task signature

luipir:
little fix to allow plugin building using invoke. Tasks help weren't aligned to tasks signatures


luipir:
this PR allow rebuild the plugin with all dependencies and submodules. Updated readme to allow correct rebuild.

PR #351, opened 2021-02-03
updates gee datasets to 2020 and fixes bug with reprojection for lpks plots

gdaldegan:


gee/landpks/src/main.py
```diff
@@ -366,7 +366,7 @@ def base_image(year, geojson, lang, gc_client, metadata):
 
     # Reproject L8 image so it can retrieve data from every latitute 
     # without deforming the aoi bounding box
-    l8sr_mosaic_bands = l8sr_y.reproject(scale=30, crs='EPSG:3857')
+    map_l8sr_mosaic = map_l8sr_mosaic.reproject(**({'crs':'EPSG:3857','scale':30}))
```
azvoleff, line 369: There is no need for the ** notation here to set the parameters from a dictionary (it is kind of a holdover from Javascript). In Python this line can just be:

`map_l8sr_mosaic = map_l8sr_mosaic.reproject(crs='EPSG:3857', scale=30)`

gee/landpks/src/main.py
```diff
@@ -418,7 +418,7 @@ def greenness(year, geojson, lang, gc_client, metadata):
     
     # Reproject ndvi mean image so it can retrieve data from every latitute 
     # without deforming the aoi bounding box
-    ndvi_mean_reproject = ndvi_mean.reproject(scale=30, crs='EPSG:3857')
+    map_mean_mosaic = map_mean_mosaic.reproject(**({'crs':'EPSG:3857','scale':30}))
```
azvoleff, line 421: Same comment here - can you remove the ** notation?

gee/landpks/src/main.py
```diff
@@ -479,7 +479,7 @@ def greenness_trend(year_start, year_end, geojson, lang, gc_client, metadata):
             .visualize(**{'palette': ['black'], 'opacity': 1})]) \
             .mosaic()
     
-    # Reproject ndvi mean image so it can retrieve data from every latitute 
+    # Reproject ndvi mean image so it can retrieve data from every latitude 
     # without deforming the aoi bounding box
     map_trnd_mosaic = map_trnd_mosaic.reproject(**({'crs':'EPSG:3857','scale':30}))
```
azvoleff, line 484: And I hadn't noticed it before, but can you fix this ** notation also?

azvoleff (changes requested):
Just a few quick changes to make the code more in line with the other GEE code (there are a few things that are more reflective of JavaScript than of Python) - can you fix these and then submit the updated pull request?

gee/landpks/src/main.py
gdaldegan: kwarg notations removed for all reprojections

Issue #341, opened 2021-01-05
Credentials should not be stored in plain text in .ini file


timlinux:
When reviewing the changes made by trends.earth to the QGIS ini file (e.g. ```~/.local/share/QGIS/QGIS3/profiles/Kartoza/QGIS```) we can find the API credentials stored in plain text:

![image](https://user-images.githubusercontent.com/178003/103630804-19886a00-4f3a-11eb-947a-56fc9a2d6beb.png)

It is best practice to use the QGIS authentication framework to do this. You can find a nice worked example of doing this in the Mergin plugin here:

https://github.com/lutraconsulting/qgis-mergin-plugin/blob/master/Mergin/utils.py#L54

luipir:
seems not so hard to follow https://docs.qgis.org/3.16/en/docs/pyqgis_developer_cookbook/authentication.html#adapt-plugins-to-use-authentication-infrastructure (well I wrote is ;) ) using NAM instead of Requests.
BTW now lastest QGIS has some QNAM classes with blocking and not bloking capability (NAM was only blocking). I should check how to integrate.
@ricardogsilva The use of creadential management have to be evaluated because would change some authentication steps repect the original plugin. Not sure if it's the focus of the project.
IMHO it's an important aspect especially for plugins proposed by public institutions.

azvoleff:
For sure - thanks for flagging, its been on the list to change this. Currently the credentials are only used to tie users to their submitted tasks on our API so as to get the URLs for the results of their tasks (the output of which are stored in a public bucket), and the passwords are random ones generated by our system, so it hasn't been a big security issue thus far.

But moving forward we've been looking towards using signed URLs to limit access to results, and better securing the credentials will be increasingly important. Once this is implemented I suppose we'll also need to request users to change their passwords given they were previously stored in plain-text, so we'll need to keep that in mind.

PR #362, opened 2021-02-11
Refactored settings GUI and logic to store credential into QGIS auth infrastructure. Fixes #354 #341

### LINKED ISSUE #354: Redesign user and authentication management
According with the implementation plan, the Settings section shall be redesigned to look like the below mock-up

geo-ldn-issue-settings-auth

Please also note the new **test connection** button, which did not feature in the original design

Implement the section that deals with user and authentication management so that both the GUI and the underlying code reflect the above vision.

### Notes

- Authentication credentials shall make use of QGIS authentication infrastrcuture
- All remaining settings shall be stored in `QgsSettings`. Use `trends_earth` as the base setting key
- `Register new user...` button shall reuse the already existing dialogue
- `Manage user account...` button shall lead to a new dialogue that combines the features currently shown by both the `Update account` and `Reset password` dialogues
- `Test connection` button shall allow the user to perform a quick connection to the GEE server to verify if the credentials are correct - Please use a [QgsMessageBar](https://docs.qgis.org/3.16/en/docs/pyqgis_developer_cookbook/communicating.html#showing-messages-the-qgsmessagebar-class) attached to the settings dialogue for providing feedback to the user on the status of the connection test

luipir:
Refactored Settings to store credentials into QGIS auth infrastructure with minimal changes in original code.

Fixes #354 #341

TODO: decide what to do with "change user" functionality. In this moment current Auth config is sores in .ini under trends_earth section. In case changing auth config this value should be changed.

### DIFF
--- a/LDMP/api.py
+++ b/LDMP/api.py
@@ -35,6 +35,7 @@ from qgis.core import (
 from LDMP.worker import AbstractWorker, start_worker
 
 from LDMP import log, debug_on
+from LDMP.message_bar import MessageBar
 
 API_URL = 'https://api.trends.earth'
 TIMEOUT = 20
@@ -71,12 +72,12 @@ def init_auth_config(email=None):
     if not previousAuthExist:
         # store a new auth config
         if not QgsApplication.authManager().storeAuthenticationConfig(currentAuthConfig):
-            iface.messageBar().pushCritical('Trends.Earth', tr_api.tr('Cannot init auth configuration'))
+            MessageBar().get().pushCritical('Trends.Earth', tr_api.tr('Cannot init auth configuration'))
             return None
     else:
         # update existing
          if not QgsApplication.authManager().updateAuthenticationConfig(currentAuthConfig):
-            iface.messageBar().pushCritical('Trends.Earth', tr_api.tr('Cannot update auth configuration'))
+            MessageBar().get().pushCritical('Trends.Earth', tr_api.tr('Cannot update auth configuration'))
             return None
 
     QSettings().setValue("trends_earth/authId", currentAuthConfig.id())
@@ -85,12 +86,12 @@ def init_auth_config(email=None):
 def remove_current_auth_config():
     authConfigId = QSettings().value("trends_earth/authId", None)
     if not authConfigId:
-        iface.messageBar().pushCritical('Trends.Earth', tr_api.tr('No authentication set. Do it in Trends.Earth settings'))
+        MessageBar().get().pushCritical('Trends.Earth', tr_api.tr('No authentication set. Do it in Trends.Earth settings'))
         return None
     log('remove_current_auth_config with authId {}'.format(authConfigId))
 
     if not QgsApplication.authManager().removeAuthenticationConfig( authConfigId ):
-        iface.messageBar().pushCritical('Trends.Earth', tr_api.tr('Cannot remove auth configuration with id: {}').format(authConfigId))
+        MessageBar().get().pushCritical('Trends.Earth', tr_api.tr('Cannot remove auth configuration with id: {}').format(authConfigId))
         return False
 
     QSettings().setValue("trends_earth/authId", None)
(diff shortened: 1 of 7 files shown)


luipir:
@Samweli probably I did a mistake creating the PR without adding the .ui refactoring but only the code... reviewing it right now

LDMP/settings.py
```diff
@@ -76,13 +91,40 @@ def __init__(self, parent=None):
 
         self.buttonBox.accepted.connect(self.close)
 
+        # load gui default value from settings
+        self.reloadAuthConfigurations()
+
+    def reloadAuthConfigurations(self):
+        self.authConfigSelect_authentication.populateConfigSelector()
```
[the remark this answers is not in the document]
luipir: yeah sorry... I did a great miskate fogottging to include .ui in the PR... rebuild it again. please hold on

LDMP/api.py
```diff
@@ -27,22 +27,119 @@
...
+        return None
+
+    # check if auth method is the only supported for no
+    if authConfig.method() != 'Basic':
+        if warn:
+            iface.messageBar().pushCritical('Trends.Earth',
+                tr_api.tr('Auth method with authId: {} is {}. Only basic auth is supported by Trend.Earth').format( authConfigId, authConfig.method() ))
+        return None
+    
+    return authConfig
```
Samweli, line 134: @luipir I thought this kind of implementation was already available out of the box from the QGIS Authentication system?

luipir: ??? what are you referring about?

LDMP/settings.py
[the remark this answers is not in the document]
luipir: now you can update the PR, I added the .ui

LDMP/api.py
Samweli: the whole process of CRUD for auth config ids.

LDMP/api.py
```diff
@@ -27,22 +27,119 @@
...
+    currentAuthConfig.setMethod('Basic')
+    currentAuthConfig.setUri(API_URL + '/auth')
+    currentAuthConfig.setConfig('username', email)
+    currentAuthConfig.setConfig('password', None)
+    currentAuthConfig.setConfig('realm', None)
+
+    if not previousAuthExist:
+        # store a new auth config
+        if not QgsApplication.authManager().storeAuthenticationConfig(currentAuthConfig):
+            iface.messageBar().pushCritical('Trends.Earth', tr_api.tr('Cannot init auth configuration'))
```
Samweli, line 74: Can we use a message bar inside plugin UI instead of the iface one.

LDMP/api.py
```diff
@@ -27,22 +27,119 @@
...
+            return None
+    else:
+        # update existing
+         if not QgsApplication.authManager().updateAuthenticationConfig(currentAuthConfig):
+            iface.messageBar().pushCritical('Trends.Earth', tr_api.tr('Cannot update auth configuration'))
```
Samweli, line 79: see above

LDMP/api.py
luipir: For me ok... may we ask opinion to @ricardogsilva just to verify in what direction to go?
BTW are you referring to message bar inside QgsAuthSelect GUI or another message bar to be added in Setting GUI?

LDMP/api.py
luipir: In this case Auth system can manage more thatn basic, but the PR try to leave the api.py and server side interacion as before. and that only support basic.
I can just only remove validity here https://github.com/ConservationInternational/trends.earth/pull/362/files#diff-cf2d71fe8784cbc1c711c47d1ff2d1d41357cf9ad2ff7f3f937d55a71ccd6480R122 
Or I misunderstood your question?

LDMP/api.py
ricardogsilva: I agree with @Samweli that messages related to something that is going on inside some extra dialogue that is currently being used by the user should preferably reported in a message bar of that dialogue - unless that is painful to implement for now.

luipir: ok, so we should add a general message bar for the DlgSettings ora message bar for each subsection of the DlgSettings or just limit auth messages using the already available messagebar inside QgsAuthSelect ?

luipir: IMHO a general message bar for DlgSettings dialog would be enough,

ricardogsilva: yes, I agree. Go with your best judgement @luipir

luipir: ok, doing it

luipir: @Samweli I'm adding a message bar inside ther DlgSettings not in the plugin docker GUI because settings is activating indipendently by the docked window. Is it ok for you?.
Any sub windows in DlgSettings should have a method to call kind of parent.logMessage
let me know... in the meantime I will proceed adding the message bar to the dialog

luipir: @ricardogsilva and @Samweli I've added a new singleton messagebar class that attach a message to a dialog when it is shown and reset it to general message bar when the dialog is hided.
In this way any API or Worker will show messages in the current Dialog (when set) instead of to the general one.

luipir: p.s for sure can be wrote better and probably can be simplified the interface to the singleton messagebar... any suggestion is welcome!

Samweli: > @Samweli I'm adding a message bar inside ther DlgSettings not in the plugin docker GUI because settings is activating indipendently by the docked window. Is it ok for you?

yes @luipir that is okay with me

LDMP/api.py
Samweli: I was concerned about the defined methods init_auth_config, remove_current_auth_config,  get_user_email and get_auth_config. I thought when dealing the QGIS Auth system, we won't have to define all these functionality and we could achieve the same using the QGIS Auth API for adding, updating and deleting auth_config details including its ids(may be my understanding about the Auth API is not correct).

luipir: The general idea is to avoid to change actual code flow of the original code... e.g. minimal modifications. so:
1) init_auth_config is the entry point when clicking register... need to create a new (or modify an existing default) auth config, 
the point directly to the new config to facilitate the user (that is done outside apy.py)
2) the same does remove_current_auth_config that is called as before in the same code flow as before when removing a user
3) get_auth_config is a wrapper to avoid to modify current code flow... done just before api.py try to fill requests with PWD
4) get_user_email is as before but using auth manager

e.g. all is using QgsAuthManager but these wrapper try to minimize changes 

I hope is more clear now

Samweli: yes thanks @luipir

ricardogsilva:
@luipir 

Did an extra commit to this PR to basically get rid of the extra dialogue for managing the user profile. It was a dialogue with only three buttons inside, so I just put the buttons on the main settings dialogue:

geo-ldn-trends-earth-settings-auth

I also did some minor tweaks to some of the dialogues (add spacers, modify widgets sizing policies, etc)

PR #363, opened 2021-02-12
Change location of region configuration into settings dialogue

### LINKED ISSUE #358: Implement controls for selecting Region of Interest in the Settings dialogue
According with the implementation plan, the Settings section shall be redesigned to look like the below mock-up

trends-earth-settings-dialogue-redesigned

Implement the section that deals with selecting the Region of Interest for executing algorithms, so that both the GUI and the underlying code reflect the above vision.

### Notes

- As shown on the above mockup, place the region of interest settings in a group. Although not shown above, it may be best to use a collapsible group
- Use the same set of widgets that already exist for storing the region of interest. These are currently shown as an additional tab in each algorithm's execution dialogue. Put them in the settings dialogue instead. This means the widgets for choosing a region of interest shall no longer be part of each algorithm's execution dialogue, they shall rather be only present in the settings dialogue.
- Store all region of interest settings in `QgsSettings`. Use `trends_earth` as the base setting key
- Make sure to adapt the current code to continue being functional with this new approach. This means that the various algorithms shall be able to get the region of interest from its new location and continue working as before

Samweli:
Fixes #358

Using the mockup in https://github.com/ConservationInternational/trends.earth/issues/358 I have added a scroll area because the region content were making the setting dialogue height so large that It could fit screen.

Removed the region/point configuration in the algorithm now they are only available via Settings dialogue.
Screenshot
latest_settings

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ invoke.yaml
 trends.earth_admin_user_credentials.json
 trends.earth_test_user_credentials.json
 .ipynb_checkpoints
+.idea/
--- a/LDMP/__init__.py
+++ b/LDMP/__init__.py
@@ -108,6 +108,7 @@ def binaries_available():
 def debug_on():
     QSettings().value("LDMP/debug", False) == 'True'
 
+
 def openFolder(path):
     if not path:
         return
--- a/LDMP/calculate.py
+++ b/LDMP/calculate.py
@@ -852,7 +852,7 @@ class DlgCalculateBase(QtWidgets.QDialog):
         if self.settings.value("trends_earth/region_of_interest/chosen_method") == 'country_region' or \
                 self.settings.value("trends_earth/region_of_interest/chosen_method") == 'country_city':
             if self.settings.value("trends_earth/region_of_interest/chosen_method") == 'country_city':
-                if not self.settings.value("trends_earth/region_of_interest/buffer_checked"):
+                if self.settings.value("trends_earth/region_of_interest/buffer_checked") != 'True':
                     QtWidgets.QMessageBox.critical(None,tr_calculate.tr("Error"),
                            tr_calculate.tr("You have chosen to run calculations for a city."
                                             "You must select a buffer distance to define the "
--- a/LDMP/calculate.py
+++ b/LDMP/calculate.py
@@ -28,7 +28,8 @@ from qgis.PyQt.QtCore import (QTextCodec, QSettings, pyqtSignal,
 from qgis.core import (QgsFeature, QgsPointXY, QgsGeometry, QgsJsonUtils,
     QgsVectorLayer, QgsCoordinateTransform, QgsCoordinateReferenceSystem,
     Qgis, QgsProject, QgsLayerTreeGroup, QgsLayerTreeLayer,
-    QgsVectorFileWriter, QgsFields, QgsWkbTypes)
+    QgsVectorFileWriter, QgsFields, QgsWkbTypes,
+    QgsSettings)
 from qgis.utils import iface
 from qgis.gui import QgsMapToolEmitPoint, QgsMapToolPan
 
@@ -645,6 +646,7 @@ class CalculationOptionsWidget(QtWidgets.QWidget, Ui_WidgetCalculationOptions):
             self.where_to_run_enabled = False
             self.groupBox_where_to_run.hide()
 
+
 class CalculationOutputWidget(QtWidgets.QWidget, Ui_WidgetCalculationOutput):
     def __init__(self, suffixes, subclass_name, parent=None):
         super(CalculationOutputWidget, self).__init__(parent)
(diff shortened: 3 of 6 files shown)


ricardogsilva (changes requested):
Some additional changes are needed.


- GUI does not adjust when I resize the settings dialogue - seems like a layout is missing? Check the screenshot below for an example:

geo-ldn-settings-dialogue-redesign-incorrect-resizing

- Selecting a point from canvas as the coordinates for the point throws an error for a missing `log` function

- When clicking the `Step2: Enter login` button I get this error:

  ```python
  An error has occurred while executing Python code: 

  NameError: name 'settings' is not defined 
  Traceback (most recent call last):
    File "/home/ricardo/.local/share/QGIS/QGIS3/profiles/trends-earth-dev/python/plugins/LDMP/settings.py", line 398, in showEvent
      password = settings.value("LDMP/password", None)
  NameError: name 'settings' is not defined
  ```

  Somehow I am still able to get the dialogue after seeing this error though. I can enter my credentials and I see another round of the same error, but apparently it works.

- be more consistent with the settings keys. Avoid mixing `snake_case` with `camelCase`. Also, don't refer to `AreaWidget` in the settings, choose a name that refers to the actual settings being stored, not the GUI widget where they are used. The layout of the settings does not need to match the GUI. It should be logical and compact. Here is my suggestion:

  ```
  trends_earth/
    region_of_interest/
      chosen_method: { country_region | country_city | point | vector_layer }
      country/
        country_name: { country-name }
        region_name: { region-name }
        city_name: { city-name }
      point/
        x: { x-coodinate }
        y: { y-coodinate }
      vector_layer: { file-path }
      buffer_size: { buffer-size-possibly-zero }
  ```

- It is not clear to me what is the purpose of these settings:

  ```
  trends_earth/AreaWidget/checkBox_custom_crs_wrap
  trends_earth/AreaWidget/custom_crs_enabled
  ```

  Are these used internally?

Samweli:
> * Are these used internally?

I think there might be some cases where they can be used, but so far I haven't found if there is any usage of them, as their corresponding GUI components are hidden when loading the Settings dialogue.

ricardogsilva (changes requested):
IMO the GUI is looking too cramped:

geo-ldn-trends-earth-settings-scroll1

Take some inspiration from the main QGIS settings dialogue and:

- Make the scroll area be the primary container for the whole widget. This means everything **except** the OK/Cancel buttons shall be scrollable
- There are too many boxes in the Region Settings section, which causes too much visual disruption. Work on the layout of the widgets so that they do not require such heavy grouping

azvoleff:
@Samweli - re `trends_earth/AreaWidget/checkBox_custom_crs_wrap` and `trends_earth/AreaWidget/custom_crs_enabled`, neither is used at present in the plugin. In past versions we used them in order to support running analyses in the cloud in a CRS other than wgs84, but then we pulled it as having flexibility in selecting a coordinate system led to too much complexity for many users. They were left in the code but not exposed in case we chose to use them in the future. At this point, fine to pull them as it is unlikely we'll implement that option in the near future.

As an aside the "custom_crs_wrap" option related to whether or not an AOI crosses the 180th meridian. If checked it would indicate that the chosen CRS could handle that crossing without needing to split the polygon in two. Given we don't have custom CRS supported at present, fine to also remove that checkbox.

Samweli:
> IMO the GUI is looking too cramped

right @ricardogsilva.

Samweli:
> Given we don't have custom CRS supported at present, fine to also remove that checkbox.

@azvoleff thanks, will update the PR with the changes.

azvoleff:
Also two other comments:

- users are likely to need to access this page fairly frequently in order to change the area of interest, as commonly users will, for example, run different areas within the same country to compare summary statistics, or to compare different sites where they made interventions. User credentials however I wouldn't see needing to be changed very often. Given that, can we move the area of interest section to the top, and have the authentication settings and advanced settings at the bottom?
- some algorithms that run in the cloud have a maximum size that they can be applied to (as some analyses are very intensive and can take hours to days to run - SDG 11.3.1 is one example where we limit the size of the AOI). Right now those checks happen when a user hits "calculate" on the relevant algorithm. A pop-up message then appears telling them to reduce the size of the AOI if it is too big. How would you envision those checks being implemented (or more importantly how would they easily redefine their AOI from the algorithm page) if the AOI selection is centralized in one place?
- On a related note, @MLNoon pointed out after the call earlier today that while having the AOI page come up multiple times across tools can be cumbersome, it does also force users to remember the area they are looking at, and to consider (hopefully!) whether that area makes sense for a given analysis. The restoration tool we've implemented for example could be run at a national scale, but doing so doesn't really make sense given the methods we are using. I think we can address these concerns in the algorithm overview sections, but just a thing to keep in mind.

ricardogsilva (changes requested):
I have a number of issues with the look of the settings dialogue. The proposed implementation looks like this:

geo-ldn-settngs-roi1

I've marked some areas which should be tweaked:

1. It looks strange that only the Region Of Interest is collapsible. Either all three groups should be collapsible (i.e. Region of interest, trends.earth remote server and Advanced) or they should all be non-collapsible - Personally I think the non-collapsible option is preferred, but I leave that to your judgement. In any case, please make them all behave the same way

2. The scroll bar looks very weird. This seems to be caused by the fact that the widgets does not have a visible frame but it still has some padding. Remove the padding on the parent widget's layout and it will probably look better. We want a similar effect as the one found on the main QGIS settings dialogue, where there is no space between the scrollbar and the window border

3. Please get rid of the additional groupbox that is surronding the region/City controls. It makes the dialogue look busier

4. Lets replace the `x` and `y` labels with `longitude` and `latitude`  and lets also make sure we always store the point in EPSG:4326

5. The buffer option is too close to the previous one, which makes the GUI look very cramped. Please give it a bit more space in there to let it breathe. Also, please also remove that additional group box that contains the controls for selecting the buffer size

For 1. if you decide that you prefer that the three sections be collapsible, take care to test how the GUI looks like when they are actually collapsed by the user. This is how it currently looks like:

geo-ldn-settngs-roi2

Please take note (and fix) the following:

1. Too much wasted space when the group box is collapsed
2. This is caused by 1 too, but the remaining sections look odd, floating in the middle of the dialogue

ricardogsilva (approved):
Looking godd now, thanks a lot!

PR #370, opened 2021-02-24
Added toggable main dock widget. Fixes #366

### LINKED ISSUE #366: Add plugin dock UI
As discussed previously, we are introducing a dock area as the main plugin widget. 

geo-ldn-plugin-dock

As seen in the mockup above, the dock shall use a tab widget with two tabs, named _Algorithms_ and _Datasets_. This issue deals only with the creation of the dock, tab widget and the two tab pages. The contents of each tab are not in scope here (they shall be handled in other issues)

### Notes
- The dock shall be enabled/disabled by the user pressing the main plugin icon (make it checkable)
- It shall be put on the right side of the main QGIS UI

luipir:


LDMP/gui/WidgetMain.ui
```diff
@@ -0,0 +1,53 @@
...
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>380</width>
+    <height>601</height>
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>Trend.Earth</string>
```
ricardogsilva: Should be Trends.Earth instead

```suggestion
   <string>Trends.Earth</string>
```

ricardogsilva (approved):
Looks good, thanks.


I did an additional commit just to change some `trend.earth` stuff to `trends.earth`. Please in the future try to be more mindful of this spelling.

Issue #368, opened 2021-02-22
Add a name to region of interest


ricardogsilva:
Enhance the UI of the settings dialogue in order to allow specifying a name to the Region of Interest (ROI) settings. This name shall be shown on the algorithm execution dialogue in order to remind the user of the current setting.

The user shall not be required to actually input a name of the ROI as this shall be computed dynamically from the current settings. The user is able to override this computed default name in order to supply a more human-friendly name. 

The chosen name must also be saved in `QgsSettings`, together with the other ROI properties.

Examples of computed names:

1. User selects a country and region as the ROI -> `{sluggified-country-name}-{sluggified-region-name}`. Ex: `portugal-setubal`
1. User selects a country and city as the ROI -> `{sluggified-country-name}_{sluggified-city-name}`. Ex: `italy-rome`
1. User selects a point as the ROI -> `pt-lon{longitude}lat{latitude}`. Ex: `pt-lon-19.345lat-22.444
1. User selects a loaded shape as the ROI -> `shape-lon{centroid-longitude}lat{centroid-latitude}`. Ex: `shape-lon21.567-lat-176.123`

#### Notes
- In the examples above, `sluggified` means that the name is modified to be safer to use in file paths, etc. Example: `United Kingdom` becomes `united-kingdom`
- If the user is also using a `buffer` around the ROI, this shall be denoted in the computed name by appending `-buffer-{buffer-length}`. This means that example 1 above would like this instead: `portugal-setubal-buffer-20`

azvoleff:
If there is room in the dialog, I'd suggest there be an option for the user to manually enter a name. While I like automatically generating a name for the region, I worry that for a user using a point or shape to define a region, the lat/lon autogenerated names won't have much meaning - a common current use case is where a national-level policymaker uses the tool to process data for their country, but needs to use their own shapefile to define the national boundary in order to ensure they are using their official country polygon. In that case it would make sense for them to be able to manually enter the country name so they could easily identify the region used.

Or another option could be to use the filename for the region name if the user uses a shape from a file.

ricardogsilva:
@azvoleff 

> I'd suggest there be an option for the user to manually enter a name

Maybe the issue description is not very clear, but that is exactly what we plan to do. We generate a default name automatically and place that in an editable text box. The user is then free to leave the default or modify it to something easier to memorize, if desired.

> Or another option could be to use the filename for the region name if the user uses a shape from a file.
[...]

PR #377, opened 2021-03-03
add name for region settings

Samweli:
Fixes https://github.com/ConservationInternational/trends.earth/issues/368

Screenshot
settings_name

### DIFF
--- a/LDMP/gui/WidgetSelectArea.ui
+++ b/LDMP/gui/WidgetSelectArea.ui
@@ -43,23 +43,8 @@
      <property name="title">
       <string>Algorithm region of interest</string>
      </property>
-     <layout class="QGridLayout" name="area_groupbox_gridLayout">
-      <property name="leftMargin">
-       <number>9</number>
-      </property>
-      <property name="topMargin">
-       <number>0</number>
-      </property>
-      <property name="rightMargin">
-       <number>0</number>
-      </property>
-      <property name="bottomMargin">
-       <number>9</number>
-      </property>
-      <property name="spacing">
-       <number>6</number>
-      </property>
-      <item row="0" column="0">
+     <layout class="QVBoxLayout" name="verticalLayout_4">
+      <item>
        <widget class="QRadioButton" name="area_fromadmin">
         <property name="text">
          <string>Country / Region</string>
@@ -72,17 +57,7 @@
         </attribute>
        </widget>
       </item>
-      <item row="11" column="0">
-       <widget class="QRadioButton" name="area_frompoint">
-        <property name="text">
-         <string>Point</string>
-        </property>
-        <attribute name="buttonGroup">
-         <string notr="true">area_radio_buttonGroup</string>
-        </attribute>
-       </widget>
-      </item>
-      <item row="1" column="0">
+      <item>
        <widget class="QFrame" name="frame_country_region">
         <property name="frameShape">
          <enum>QFrame::NoFrame</enum>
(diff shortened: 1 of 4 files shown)


luipir:
ok, I'll give a look as soon as possible @Samweli

luipir:
@Samweli just a quick though seeing the animated gif... if the name have to be used as folder or file name, what else to avoid to use "(" or other "." other that used in file extension?

LDMP/settings.py
```diff
@@ -235,6 +238,15 @@ def __init__(self, parent=None):
         self.populate_cities()
         self.area_admin_0.currentIndexChanged.connect(self.populate_admin_1)
         self.area_admin_0.currentIndexChanged.connect(self.populate_cities)
+        self.area_admin_0.currentIndexChanged.connect(self.generate_name_setting)
+
+        self.secondLevel_area_admin_1.currentIndexChanged.connect(self.generate_name_setting)
```
luipir, line 243: what would be the meaning of "admin_1" in secondLevel_area_admin_1 combobox menu?

LDMP/settings.py
```diff
@@ -301,6 +313,9 @@ def load_settings(self):
         if buffer_size:
             self.buffer_size_km.setValue(float(buffer_size))
         self.checkbox_buffer.setChecked(buffer_checked)
+        self.area_settings_name.setText(
+            self.settings.value("trends_earth/region_of_interest/area_settings_name", '')
```
luipir, line 317: as far as I can see  there is a principal problem in this code is that if a use set a custom name, any area parameter modification will override custom string.
Would it make sense to add a checkbox locking the generated or custom string?

luipir:
@Samweli other the aspect to lock the area name, IMHO LGTM. Need to have opinion by @ricardogsilva about a locking checkbox and eventually removing some special character in ara name

ricardogsilva:
I think it is OK to let the code regenerate the custom name whenever the user changes some area parameter for now.

ricardogsilva:
@luipir 

if you are :+1: on this PR, please proceed to merge

LDMP/gui/WidgetSelectArea.ui
```diff
@@ -522,6 +507,20 @@
         </layout>
        </widget>
       </item>
+      <item>
+       <layout class="QHBoxLayout" name="horizontalLayout_2">
+        <item>
+         <widget class="QLabel" name="label_2">
+          <property name="text">
+           <string>Settings name:</string>
```
luipir, line 515:
```suggestion
           <string>Area name:</string>
```

I'm confused about the label you used for the tag of the Area.
Wait for resolve this to merge

luipir:
@Samweli would it make sense to set the first time value in no setting is available calling one time generate_name_setting() ?
I started the setting GUI and area name is set to empty... is the desired behaviour?

luipir:
just as note and out of the scope of the PR... the area name continue to have the same localization problem as the region/city giving back a strange string.

luipir (changes requested):

Samweli:
> @Samweli just a quick though seeing the animated gif... if the name have to be used as folder or file name, what else to avoid to use "(" or other "." other that used in file extension?

@luipir we are not saving the area name as a file name or folder, so I guess we won't have to do a check for those characters.

LDMP/settings.py
Samweli: I'm just reusing the widgets name as defined from the area widget. Do you want us to update them?

LDMP/settings.py
Samweli: @luipir I think that is the expected behaviour, also this is only executed once when opening up the settings dialog.

Samweli:
> @Samweli would it make sense to set the first time value in no setting is available calling one time generate_name_setting() ?
> I started the setting GUI and area name is set to empty... is the desired behaviour?

Yes I think that right @luipir, we should not set it to empty when using the dialog for the first time, @ricardogsilva what do you think?

LDMP/gui/WidgetSelectArea.ui
Samweli: ok let me change it.

LDMP/settings.py
luipir: no... it's ok

LDMP/settings.py
luipir: ok

luipir (approved):

luipir:
> Yes I think that right @luipir, we should not set it to empty when using the dialog for the first time, @ricardogsilva what do you think?

just to speedup I merge it, we can manage detail (like first name value setup) in an issue later @Samweli

Samweli:
thanks @luipir

PR #382, opened 2021-03-11
MVC implemented. Fixes #367

### LINKED ISSUE #367: Implement algorithms tab of the plugin dock
The plugin dock has an `Algorithms` tab. This is meant to be the primary place where users can discover and run Trends.Earth algorithms.

geo-ldn-dock-algorithms-tab

As seen on the mockup above, the algorithms section shows a list of available algorithms which can be used for generating multiple SDG-related indicators. The list is organized with a three-level hierarchical structure:
- The **first level is the group** of each algorithm
- The **second level has the name of the algorithm and a button to execute it**. Depending on the algorithm, the Run button may be compound, with a main action that allows running with custom data. The secondary action shall allow running with the default UNCCD datasets. Algorithms that can only run using remote data shall show a normal button. This shall serve as a hint to the user that using own data is now the preferred execution method
- The **third level has a brief description** about the algorithm and a link to more information

Algorithms shall primarily be grouped in accordance with their respective SDG.

There may also be other groups, to show beta algorithms or some other related categories. These shall be shown last on the list, in order to indicate to the user that their relevance is lower

Clicking on one of the Run buttons shall open an additional dialogue that allows parametrizing and running the relevant algorithm (which will be the subject of additional issues)

### Notes

- Some algorithms may execute only locally (with the user's own datasets), some may execute only remotely (using Trends.Earth default datasets), some may be able to execute either locally or remotely, depending on the user's choice. This information shall be conveyed in the GUI by doing the following:
	- If the algorithm only runs locally or only runs remotely, the algorithm list shall show a normal `Run` button that will cause its algorithm execution dialogue to open
	- If the algorithm may run either locally or remotely, the algorithm list shall show a compound button, with the primary action being `Run` to run the algorithm locally. The secondary action shall be `Run with default data` and it will allow running the algorithm remotely

- Lets introduce a class to model algorithms groups. This shall have at least the following properties:
	- `name: str` - Human friendly name of the group, to be shown on the GUI. Example: `Proportion of land that is degraded over total land area`
	- `name_detail: typing.Optional[str]` - An optional further detail about the group, to be shown on the GUI after the name, enclosed in brackets. Example: `SDG 15.3.1` 
	
- Lets also introduce (or improve upon, if it is there already) a class to model algorithms. This shall have at least the following properties:
	- `group: AlgorithmGroup` - group that the algorithm belongs to. Shall be an instance of the class described in the previous point
	- `runs_locally: bool` - boolean that specifies that the algorithm can be run locally, with user supplied datasets. This shall be used by the GUI in order to build the algorithm's run action
	- `runs_remotely: bool` - boolean that specifies that the algorithm can be run remotely, with Trends.Earth default datasets. This shall be used by the GUI in order to build the algorithm's run action
	- `name: str` - Human friendly name of the algorithm, to be shown on the GUI. Example: `Land Productivity`
	- `name_detail: typing.Optional[str]` - An optional further detail about the name of the algorithm that will be shown on the GUI after the name, enclosed in brackets. Example: `sub-indicator 1`
	- `brief_description: str` - Brief description of the algorithm. One or two paragraphs of text max. This shall be shown as the third level of the algorithms list on the GUI.
	- `description: str` - Longer description about the algorithm, with enough text to help the user configure the algorithm parameters in order to produce suitable results. This shall be shown on the algorithm execution dialogue.

luipir:
Implemented MVC to show algorithms with custom delegate to trigger run actions

TODO:
1) Groups has a basic moving pixmap. better would be using SVG animation => need more investigation and test to do it at work with QSvgPainter or something equivalent.
2) Need a better graphic organization... UX is too basic and need more work for better representation of delegated widget
3) For now algorithms are fixed as test data. Need a way to load algorithm list from a conf file and setup model

Teh code is flexible enough to allow:
1) discriminate editing basing on row/column index
2) can create a custom widget editor (and rendered when not in editing) for each row/column index

all modification should be limited (almost completely) inside the algorithm view delegate.

fixes #367

### DIFF
--- a/LDMP/main_widget.py
+++ b/LDMP/main_widget.py
@@ -39,7 +39,7 @@ class MainWidget(QtWidgets.QDockWidget, Ui_dockWidget_trends_earth):
         super(MainWidget, self).__init__(parent)
 
         self.setupUi(self)
-        self.setupAlgorithsTree()
+        self.setupAlgorithmsTree()
 
     def setupAlgorithmsTree(self):
         # setup algorithms and their hierarchy
--- a/LDMP/main_widget.py
+++ b/LDMP/main_widget.py
@@ -18,6 +18,9 @@ from qgis.core import QgsSettings
 from LDMP import __version__, log
 from LDMP.message_bar import MessageBar
 from LDMP.gui.WidgetMain import Ui_dockWidget_trends_earth
+from LDMP.models.algorithms import AlgorithmGroup, AlgorithmDescriptor, AlgorithmRunMode
+from LDMP.models.algorithms_model import AlgorithmTreeModel
+from LDMP.models.algorithms_delegate import AlgorithmItemDelegate
 
 settings = QgsSettings()
 
@@ -36,6 +39,36 @@ class MainWidget(QtWidgets.QDockWidget, Ui_dockWidget_trends_earth):
         super(MainWidget, self).__init__(parent)
 
         self.setupUi(self)
+        self.setupAlgorithsTree()
+
+    def setupAlgorithsTree(self):
+        # setup algoriths and it's hierarchy
+        tree = AlgorithmGroup(name='root', name_details='root detauils', group=None, algorithms=[])
+        first = AlgorithmGroup(name='firstG', name_details='first details', group=tree, algorithms=[])
+        second = AlgorithmGroup(name='secondG', name_details='second details', group=tree, algorithms=[])
+
+        alg1G1 = AlgorithmDescriptor(name='alg1G1', name_details='alg1G1 details', brief_description='alg1G1 brief descr', description='alg1G1 long description', group=first)
+        alg1G1.run_mode = AlgorithmRunMode.Both
+        alg2G1 = AlgorithmDescriptor(name='alg2G1', name_details='alg2G1 details', brief_description='alg2G1 brief descr', description='alg2G1 long description', group=first)
+        first.algorithms.append(alg1G1)
+        first.algorithms.append(alg2G1)
+
+        alg1G2 = AlgorithmDescriptor(name='alg1G2', name_details='alg1G2 details', brief_description='alg1G2 brief descr', description='alg1G2 long description', group=second)
+        alg1G2.run_mode = AlgorithmRunMode.Both
+        second.algorithms.append(alg1G2)
+
+        tree.algorithms.append(first)
+        tree.algorithms.append(second)
+
+        # show it
+        algorithmsModel = AlgorithmTreeModel(tree)
+        self.treeView_algorithms.setModel(algorithmsModel)
+        delegate = AlgorithmItemDelegate(self.treeView_algorithms)
+        self.treeView_algorithms.setItemDelegate(delegate)
+
+        # configure View how to enter editing mode
+        self.treeView_algorithms.setEditTriggers(QtWidgets.QAbstractItemView.AllEditTriggers)
+
 
     def closeEvent(self, event):
         super(MainWidget, self).closeEvent(event)
--- a/LDMP/main_widget.py
+++ b/LDMP/main_widget.py
@@ -41,9 +41,9 @@ class MainWidget(QtWidgets.QDockWidget, Ui_dockWidget_trends_earth):
         self.setupUi(self)
         self.setupAlgorithsTree()
 
-    def setupAlgorithsTree(self):
-        # setup algoriths and it's hierarchy
-        tree = AlgorithmGroup(name='root', name_details='root detauils', group=None, algorithms=[])
+    def setupAlgorithmsTree(self):
+        # setup algorithms and their hierarchy
+        tree = AlgorithmGroup(name='root', name_details='root details', group=None, algorithms=[])
         first = AlgorithmGroup(name='firstG', name_details='first details', group=tree, algorithms=[])
         second = AlgorithmGroup(name='secondG', name_details='second details', group=tree, algorithms=[])
 
--- a/LDMP/main_widget.py
+++ b/LDMP/main_widget.py
@@ -18,7 +18,12 @@ from qgis.core import QgsSettings
 from LDMP import __version__, log
 from LDMP.message_bar import MessageBar
 from LDMP.gui.WidgetMain import Ui_dockWidget_trends_earth
-from LDMP.models.algorithms import AlgorithmGroup, AlgorithmDescriptor, AlgorithmRunMode
+from LDMP.models.algorithms import (
+    AlgorithmGroup,
+    AlgorithmDescriptor,
+    AlgorithmRunMode,
+    AlgorithmDetails
+)
 from LDMP.models.algorithms_model import AlgorithmTreeModel
 from LDMP.models.algorithms_delegate import AlgorithmItemDelegate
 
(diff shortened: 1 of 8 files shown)


luipir:
@ricardogsilva I set you as reviewer... feel free to set @Samweli basing on you personal workload and task organization.

ricardogsilva:
I'll review ASAP

LDMP/main_widget.py
```diff
@@ -36,6 +39,36 @@ def __init__(self, parent=None):
         super(MainWidget, self).__init__(parent)
 
         self.setupUi(self)
+        self.setupAlgorithsTree()
+
+    def setupAlgorithsTree(self):
+        # setup algoriths and it's hierarchy
+        tree = AlgorithmGroup(name='root', name_details='root detauils', group=None, algorithms=[])
```
ricardogsilva, line 46: Please take care of these typos

```suggestion
    def setupAlgorithmsTree(self):
        # setup algorithms and their hierarchy
        tree = AlgorithmGroup(name='root', name_details='root details', group=None, algorithms=[])
```

LDMP/models/algorithms.py
```diff
@@ -0,0 +1,138 @@
...
+
+class AlgorithmBase(abc.ABC):
+
+    def __init__(self,
+            name: str,
+            name_details: Optional[str] = None,
+            group: 'AlgorithmGroup' = None, # string because forwarde declared class
+        ) -> None:
+        super().__init__()
+        self.type: Optional[AlgorithmNodeType] = None
```
ricardogsilva, line 43: Please use another name for this property, as `type` is also the name of a function in Python. I suggest `algorithm_type`

```suggestion
        self.algorithm_type: Optional[AlgorithmNodeType] = None
```

LDMP/models/algorithms.py
```diff
@@ -0,0 +1,138 @@
...
+    @abc.abstractmethod
+    def rowCount(self) -> int:
+        pass
+
+    @abc.abstractmethod
+    def columnCount(self) -> int:
+        pass
+
+    @abc.abstractmethod
+    def child(self, row) -> Union['AlgorithmBase', None]: # string because AlgorithmBase not yet completly defined
```
ricardogsilva, line 72: A small nitpick, but I guess `Optional` is clearer than `Union` here

```suggestion
    def child(self, row) -> Optional['AlgorithmBase']:  # string because AlgorithmBase not yet completly defined
```

LDMP/models/algorithms.py
```diff
@@ -0,0 +1,138 @@
...
+        pass
+
+class AlgorithmGroup(AlgorithmBase):
+
+    def __init__(self,
+            name: str,
+            name_details: Optional[str],
+            group: 'AlgorithmGroup',
+            algorithms: List[Union['AlgorithmDescriptor', 'AlgorithmGroup']]
```
ricardogsilva, line 81: We did not design groups to be nestable. The idea is to have group -> algorithms. If it makes the implementation simpler I don't mind to keep them nestable, but if this adds additional complication I suggest to remove this possibility.

LDMP/models/algorithms.py
```diff
@@ -0,0 +1,138 @@
...
+    def rowCount(self) -> int:
+        return len(self.algorithms)
+    
+    def child(self, row: int) -> Union['AlgorithmBase', None]:
+        if row < 0 or row >= len(self.algorithms):
+            return None
+        
+        return self.algorithms[row]
+    
+    def columnName(self, column: int) -> Union[str, None]:
```
ricardogsilva, line 102: It would be cleaner to not show the column names on the GUI

LDMP/models/algorithms.py
```diff
@@ -0,0 +1,138 @@
...
+    def __init__(self,
+            name: str,
+            name_details: Optional[str],
+            brief_description: str,
+            description: str,
+            group: AlgorithmGroup, # e.g. an Alg can belogs to only a group => 1 to 1 limitation!
+            run_mode: AlgorithmRunMode = AlgorithmRunMode.Locally,
+        ) -> None:
+        super().__init__(name, name_details, group)
+        self.type = AlgorithmNodeType.Leaf
```
ricardogsilva, line 122: According to the original design, the tree is supposed to have three levels:

group -> algorithm -> description

LDMP/models/algorithms_delegate.py
```diff
@@ -0,0 +1,263 @@
...
+    ".... x    x ....",
+    "..... x  x .....",
+    "..... x  x .....",
+    ".... x ss x ....",
+    "... x ssss x ...",
+    ".. x ssssss x ..",
+    ". x          x .",
+    ". xxxxxxxxxxxx .",
+    ".              .",
+    "................"]
```
ricardogsilva, line 146: As much as I appreciate this ASCII art I don't think we should be adding icons to the algorithms tab, at least not at this early stage of development. This was not part of our original design, so we are getting into more trouble than initially foreseen with this.

LDMP/models/algorithms_delegate.py
```diff
@@ -0,0 +1,263 @@
...
+        self.pixmaps = (QPixmap(_icon0_xpm),
+                        QPixmap(_icon1_xpm),
+                        QPixmap(_icon2_xpm),
+                        QPixmap(_icon3_xpm))
+        self.needsRedraw.connect(parent.viewport().update)
+
+        # create svg renderer
+        # self.svg_renderer = QSvgRenderer('/home/ginetto/Downloads/tcjp3h.svg', parent)
+        # self.svg_renderer = QSvgRenderer('/home/ginetto/Downloads/spinner-of-dots-svgrepo-com.svg', parent)
+        # self.svg_renderer.repaintNeeded.connect(parent.viewport().update)
```
ricardogsilva, line 181: Please remove this if it is not being used, as it creates clutter

LDMP/models/algorithms_delegate.py
```diff
@@ -0,0 +1,263 @@
...
+
+        # create svg renderer
+        # self.svg_renderer = QSvgRenderer('/home/ginetto/Downloads/tcjp3h.svg', parent)
+        # self.svg_renderer = QSvgRenderer('/home/ginetto/Downloads/spinner-of-dots-svgrepo-com.svg', parent)
+        # self.svg_renderer.repaintNeeded.connect(parent.viewport().update) 
+
+    def timerEvent(self, event):
+        if event.timerId() == self.timerId:
+            self.current = (self.current + 1) % 4
+            self.needsRedraw.emit()
```
ricardogsilva, line 186: It is my understanding that this is just to update the hourglass icon. Is this correct? If so, please remove this code and related functionality in order to make this PR leaner and easier to grasp.

LDMP/models/algorithms_delegate.py
```diff
@@ -0,0 +1,263 @@
...
+    "..... x  x .....",
+    ".... x ss x ....",
+    "... x ssss x ...",
+    ".. x ssssss x ..",
+    ". x          x .",
+    ". xxxxxxxxxxxx .",
+    ".              .",
+    "................"]
+
+class AlgorithmItemDelegate(QStyledItemDelegate):
```
ricardogsilva, line 148: I'm not sure using a delegate is the best option here. Perhaps we can use [QAbstractItemView.setIndexWidget](https://doc.qt.io/qt-5/qabstractitemview.html#setIndexWidget) instead and just use a custom widget?

My main reasoning for this is that I don't like the current behavior where we need to click once on the tree row just to activate it and make the delegate be rendered in editing mode before we can then click the button for running an algorithm, which seems clumsy from a UX point of view.

Maybe there is another way to do it with a delegate, I'm not sure. In any case, what I want to prevent is the user not knowing how many times he needs to click on the button to get it to run

ricardogsilva (changes requested):
It is a great effort, but I think it needs a bit more work. Please see my inlined comments, and also this additional feedback:

- First level of indentation does not need a status icon. Indeed we did not design any type of feedback for this tab. Lets leave this type of improvement for a later stage, if at all. Please remove the hourglass icon that is currently present and all associated code.

- The third level of indentation is missing from the PR. Please check the issue description for the intended design. This seems like a substantial missing piece, since I believe it has implications on what the underlying data model shall look like. Please implement the third level.

- Please implement a more realistic set of algorithms just so that we may appreciate how things look like with some more explicit names and descriptions. This helps validating the UI spacing. Instead of:

```
firstG[first details]
  alg1G1[alg1G1 details]
...
```
use the copy shown on the issue description:
```
Proportion of land that is degrated over total land area (SDG 15.3.1)
  Land Productivity (sub-indicator 1)
```

- As mentioned inline, I'm not convinced by the usage of a custom `QStyledItemDelegate`. The UI looks a bit clumsy. I understand this is a first draft, but there are usability concerns that I'm not sure can be addressed, like:
  - the user needs to click a row once just to activate it and then he needs to click it again to push the button
  - current row looks funky because the selected highlight does not cover the full row, only a small part on the left

luipir:
> * First level of indentation does not need a status icon. Indeed we did not design any type of feedback for this tab. Lets leave this type of improvement for a later stage, if at all. Please remove the hourglass icon that is currently present and all associated code.

I agree... I mixed concept about algorithm run and data sets... btw is a good exercise when implementing datasets.

> * The third level of indentation is missing from the PR. Please check the issue description for the intended design. This seems like a substantial missing piece, since I believe it has implications on what the underlying data model shall look like. Please implement the third level.

That's a possible problem limitation... I find the QTreeView extremly flexible when sort or order elements but completly hard to adapt to multimple type in the tree... e.g. first level shoudl have 1st column only, 2nd level at least two columns and 3rd level again one column. The simple workaround is all with two column and filling necessary column at every level... but the UX representation is not so fancy.

> * Please implement a more realistic set of algorithms just so that we may appreciate how things look like with some more explicit names and descriptions. This helps validating the UI spacing. Instead of:

ok

> * As mentioned inline, I'm not convinced by the usage of a custom `QStyledItemDelegate`. The UI looks a bit clumsy. I understand this is a first draft, but there are usability concerns that I'm not sure can be addressed, like:
>   
>   * the user needs to click a row once just to activate it and then he needs to click it again to push the button
>   * current row looks funky because the selected highlight does not cover the full row, only a small part on the left

Yeh I strongly agree.. I tried to minimize this effect just using common configuration flags... using common conf dflags seems there is NO way to avoid doble click... a workaround is adding a mouse over event that start/stop editing when the mouse pass over the cell.

Another BIG BIG BIG workaround is to have yust a list/tree with only a column with a custom widget inside with collapsible group box ad a well designed widget for a single algorithm tree idem. e.g. 12st, 2nd and 3rd algorithm lever whould be in a single widget representing an alg item

LDMP/models/algorithms.py
luipir: no, it shouldn't add any complexity more

LDMP/models/algorithms.py
luipir: I leaved only for review purpose to leave clear that there is a column with description and a column for a widget... it will be hided or removed

LDMP/models/algorithms.py
luipir: ok... I'll try to implement without adding more complexity to the model

LDMP/models/algorithms_delegate.py
luipir: ok it will be removed...

LDMP/models/algorithms_delegate.py
luipir: this is placeholder just to remember that is possible to add a dynamic svg probably useful for the Dataset tree view part... please leave there temporarily to until the end of datasets tab implementation

LDMP/models/algorithms_delegate.py
luipir: ies it is... but could be also used in case of dynamic SVG... btw removing glass icon it is unuseful

LDMP/models/algorithms_delegate.py
luipir: I tried to avoid setIndexWidget because it imply low performances and should be avoided... it's not out case with few element of the tree, BTW I'll try to use setItemWidget.
I agree that the user shouldn't have to click to activate editing

ricardogsilva:
> Yeh I strongly agree.. I tried to minimize this effect just using common configuration flags... using common conf dflags seems there is NO way to avoid doble click... a workaround is adding a mouse over event that start/stop editing when the mouse pass over the cell.

> Another BIG BIG BIG workaround is to have yust a list/tree with only a column with a custom widget inside with collapsible group box ad a well designed widget for a single algorithm tree idem. e.g. 12st, 2nd and 3rd algorithm lever whould be in a single widget representing an alg item

Keep cracking at it until you (hopefully) have a solution that works :sweat_smile: . I'm fine with any implementation, as long as it feels natural to use. Once you are happy with it, please re-request a review from me.

luipir:
@ricardogsilva @Samweli this is the result of the MVC model on a real use case

Screenshot from 2021-03-18 17-53-24

as you can see in the image, first level text is not wrapped... the reason is that is almost impossible in a simple way other than using a delegated widget as is done in the II level where the text is wrapped during resize.
Before adding more code line if necessary... I would prefer a review before.

luipir:
Having a widget at any level allow us to have a heacy customizable presentation of each level... e.g. adding icons, wrapping text and so on... the model has many flexibility.
II level is a widget set in editing mode when mouse enter the cell... this allow avoiding click or double click to activate toolbutton actions

ricardogsilva (approved):
Nice work on using the persistent editor, I didn't know about it and it seems quite nice :)

LGTM!

luipir:
> Nice work on using the persistent editor, I didn't know about it and it seems quite nice :)

persistent editor is a kind of shotrcut to createEditor in StyleDelegate (or eventually EditorFactoryDelegate)

PR #385, opened 2021-03-24
Datasets tab in plugin dock. Fixes #371

### LINKED ISSUE #371: Implement datasets tab of the plugin dock
The plugin dock has a `Datasets` tab. This is meant to be the primary place where the user manages and interacts with Trends.Earth data.

geo-ldn-datasets-tab-initial

As depicted by the above mockup, the datasets tab can be split into three areas:

1. An actionable list of known datasets. 

   This list shall be scrollable, in order to accomodate a large number of entries.

   Each entry in the list represents a known dataset. It shall feature the following:

   - Dataset name
   - Creation date
   - Source of the dataset
   - Current state (available|being generated|downloading|unavailable)
   - UI controls to manage the dataset: load, details, delete

2. Some UI controls to operate on the list as a whole

   These allow filtering and ordering the list and also performing a manual refresh.

   In this context, refreshing means performing both:

   - A scan of the `base_dir` in order to discover any new datasets
   - A poll of the remote server in order to get updates on datasets that are being generated

3. Additional UI controls to import new datasets 
   
   These shall allow the user to introduce new datasets to Trends.Earth by either:

   - Downloading some sample data from the Trends.Earth servers
   - Importing local datasets

### known datasets

Known datasets are those whose metadata can be extracted by inspection of the Trends.Earth `base_dir` (as configured in the settings dialogue).

These datasets are generated by either:

1. Running one of the available Trends.Earth algorithms;
1. Downloading from the Trends.Earth sample data repositories
1. Importing from some (local) location

Each known dataset is described by an auxiliary `.json` file that carries all relevant metadata about it. These json files must be located inside the base dir in order for Trends.Earth to be able to find the respective dataset. 

Datasets can be located everywhere on the user's machine, just as long as their respective `.json` file is under the Trends.Earth base dir. This means that inside the json file there must be a property that informs the plugin of the full path to the dataset. The current implementation features a `file` property, which we can continue using, but with a full path instead of just a filename.

### Trends.Earth base dir 

The structure of the base dir shall be:

```
> {trends-earth-base-dir}/
|   v downloaded-sample-datasets/
|   |    dataset_1.json
|   |    dataset_2.json
|   |    ...
|   |    dataset_n.json
|   v imported-datasets/
|   |    dataset_1.json
|   |    dataset_2.json
|   |    ...
|   |   dataset_n.json
|   v in-transit/
|   |   dataset_1.json
|   |   dataset_2.json
|   |   ...
|   |   dataset_n.json
|   v outputs/
|   |   v {algorithm-group-1}/
|   |   |   v {algorithm-name1}/
|   |   |   |   v {execution-date-1}/
|   |   |   |   |   dataset_1.json
|   |   |   |   |   dataset_2.json
|   |   |   |   |   ...
|   |   |   |   |   dataset_n.json
|   |   |   |   > {execution-date-2}/
|   |   |   |   > ...
|   |   |   |   > {execution-date-n}/
|   |   |   > {algorithm-name-2}/
|   |   |   > ...
|   |   |   > {algorithm-name-n}/
|   |   > {algorithm-group-2}/
|   |   > ...
|   |   > {algorithm-group-n}/
```

- `downloaded-sample-datasets` - This directory is used to put all json files and all datasets that are downloaded from Trends.Earth data servers
- `imported-datasets` - This directory is used to put all json files of datasets that are imported. It may optionally also be used to copy the imported files (but we'll likely provide a UI control for the user to decide if the dataset shall be copied or not)
- `in-transit-datasets` - This directory is used to put all json files of datasets that are still being generated on a remote server. The contents of this directory are therefore volatile.
- `outputs` - All datasets and json files that are generated by a Trends.Earth algorithm shall be placed under this directory. It is further discretized according to each algorithm's group and name and execution date (as in year, month, day)

## Objective

Implement an initial version of the datasets tab that has the following:

- dataset list, with existing datasets being scanned from the base dir, as mentioned above. for now it is OK to only add:

  - dataset name
  - creation date 
  - source
  - Button to load the dataset onto QGIS canvas

  We shall add the remaining features later. Polling the remote server is alos not in scope for the current issue, we shall implement it later

- dataset import section, with two buttons (Import and download sample data), as shown in the mockup. These buttons shall reuse the existing dialogues

- list controls section. Implement only the `refresh` button for now. We will be adding list sorting and filtering later

luipir:
Screenshot from 2021-03-24 11-21-27


luipir:
@ricardogsilva houch! I forgot to check editability... tnx for the fix

ricardogsilva:
@luipir 

After reviewing I decided to push an additional commit, mostly with the following tweaks:

- Fixed tree item not being set to editable, so it was not possible to trigger the load/details/delete buttons
- Made scrolling smoother in order to avoid disorientating the user
- Added wrap to the labels that show a dataset name and source, as this would get cut off for the case of long names
- Dataset name is now shown with bold text

It seems mostly good to merge. However, I noticed a weird problem which I am leaving for you to fix. When the dataset name or source is too long the widget actually changes its size when hovered by the mouse, which causes some of the content to disappear off-screen.

Please check it out in the below GIF (which also showcases all modifications done in my commit)

Peek 2021-03-24 16-07

Please fix this issue and then I'll be +1 for merging

ricardogsilva (changes requested):
As stated in the comments, let's fix the issue with the resizing of the widgets on mouse hover before merging

luipir:
> As stated in the comment above, let's fix the issue with the resizing of the widgets on mouse hover before merging

I'm having difficulties to fix this resize aspect :/

ricardogsilva:
Thanks for the fix @luipir - some real wizard skills in there :smile:

LGTM!

luipir:
yes... the trick is the overrided method that I didn't know about editor size

PR #387, opened 2021-03-30
New list of algorithms

### LINKED ISSUE #386: Hook existing algorithm execution dialogues to the new UI components
Now that the new plugin dock is in place we need to start hooking the existing algorithms to work with it. 

- Create some sort of registry where algorithms are defined, in scope of #383 
- Load registered algorithms into their respective section of the UI. Show their description and also the relevant `Run` buttons
- Parametrize the algorithm generation and send it to be executed remotely

luipir:
New list of algorithms following 1/1 the original interface

fixes #386

### DIFF
--- a/LDMP/main_widget.py
+++ b/LDMP/main_widget.py
@@ -121,40 +121,40 @@ class MainWidget(QtWidgets.QDockWidget, Ui_dockWidget_trends_earth):
     def setupAlgorithmsTree(self):
         # setup algorithms and their hierarchy
         tree = AlgorithmGroup(name='root', name_details='root details', parent=None, algorithms=[])
-        land_degradation_group = AlgorithmGroup(name='Land degradation', name_details='SDG 15.3.1', parent=tree, algorithms=[])
-        urban_change_group = AlgorithmGroup(name='Urban change and land consumption', name_details='SDG 11.3.1', parent=tree, algorithms=[])
-        total_carbon_group = AlgorithmGroup(name='Total carbon', name_details='above and belowground, emissions and deforestation', parent=tree, algorithms=[])
-        potential_change_group = AlgorithmGroup(name='Potential change in biomass due to restoration', name_details='above and belowground woody', parent=tree, algorithms=[])
+        land_degradation_group = AlgorithmGroup(name=tr('Land degradation'), name_details=tr('SDG 15.3.1'), parent=tree, algorithms=[])
+        urban_change_group = AlgorithmGroup(name=tr('Urban change and land consumption'), name_details=tr('SDG 11.3.1'), parent=tree, algorithms=[])
+        total_carbon_group = AlgorithmGroup(name=tr('Total carbon'), name_details=tr('above and belowground, emissions and deforestation'), parent=tree, algorithms=[])
+        potential_change_group = AlgorithmGroup(name=tr('Potential change in biomass due to restoration'), name_details=tr('above and belowground woody'), parent=tree, algorithms=[])
 
         # land_degradation_group
-        land_roductivity_alg = AlgorithmDescriptor(name='Land productivity', name_details=None, brief_description=None, details=None, parent=land_degradation_group)
+        land_roductivity_alg = AlgorithmDescriptor(name=tr('Land productivity'), name_details=None, brief_description=None, details=None, parent=land_degradation_group)
         land_roductivity_alg.run_mode = AlgorithmRunMode.Remotely
-        land_roductivity_alg_details = AlgorithmDetails(name=None, name_details=None, description='TODO: Land productivity long description', parent=land_roductivity_alg)
+        land_roductivity_alg_details = AlgorithmDetails(name=None, name_details=None, description=tr('TODO: Land productivity long description'), parent=land_roductivity_alg)
         land_roductivity_alg.setDetails(land_roductivity_alg_details)
 
-        land_cover_alg = AlgorithmDescriptor(name='Land cover', name_details=None, brief_description=None, details=None, parent=land_degradation_group)
+        land_cover_alg = AlgorithmDescriptor(name=tr('Land cover'), name_details=None, brief_description=None, details=None, parent=land_degradation_group)
         land_cover_alg.run_mode = AlgorithmRunMode.Remotely
-        land_cover_alg_details = AlgorithmDetails(name=None, name_details=None, description='TODO: Land cover long description', parent=land_cover_alg)
+        land_cover_alg_details = AlgorithmDetails(name=None, name_details=None, description=tr('TODO: Land cover long description'), parent=land_cover_alg)
         land_cover_alg.setDetails(land_cover_alg_details)
 
-        soil_organic_carbon_alg = AlgorithmDescriptor(name='Soil organic carbon', name_details=None, brief_description=None, details=None, parent=land_degradation_group)
+        soil_organic_carbon_alg = AlgorithmDescriptor(name=tr('Soil organic carbon'), name_details=None, brief_description=None, details=None, parent=land_degradation_group)
         soil_organic_carbon_alg.run_mode = AlgorithmRunMode.Remotely
-        soil_organic_carbon_alg_details = AlgorithmDetails(name=None, name_details=None, description='TODO: Soil organic carbon long description', parent=soil_organic_carbon_alg)
+        soil_organic_carbon_alg_details = AlgorithmDetails(name=None, name_details=None, description=tr('TODO: Soil organic carbon long description'), parent=soil_organic_carbon_alg)
         soil_organic_carbon_alg.setDetails(soil_organic_carbon_alg_details)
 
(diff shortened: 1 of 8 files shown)


ricardogsilva:
I've taken a look at the current progress:

1. Let's not use square brackets on the short name of the algorithm and lets also swap the order of the name and short name, *i.e.* instead of `Land degradation [SDG 15.3.1]` lets do `SDG 15.3.1 - Land degradation` instead
1. Let's tuck the experimental algorithms into an `Experimental` section. This means there will only be three entries at the first level of the algorithms tree view: `SDG 15.3.1`, `SDG 11.3.1` and `Experimental` - This shall make it more clear that the experimental algorithms are not production-ready
1. The more I work with it, the more I am convinced that we should not have a third level on the tree. I'll open an issue for it, as discussed yesterday
1. It looks weird that there is a mouse hover effect on the tree level 2 and 3 but there is none on level 1. Let's get rid of that hover effect on levels 2 and 3 to make it more consistent with the rest of QGIS UI

luipir:
@ricardogsilva rebuild tree... now I'll remove the over effect
Screenshot from 2021-03-30 12-58-22

luipir:
@ricardogsilva I've trouble setting stylesheet for tree items and at moment I can't find how to remove gray background. I suggest to convert the "item background" as issue and solve later.
I would focus on adding business logic (.e.g link run to old algorithm GUIs)
Let me know if it's ok for you

ricardogsilva:
> @ricardogsilva I've trouble setting stylesheet for tree items and at moment I can't find how to remove gray background. I suggest to convert the "item background" as issue and solve later.
> I would focus on adding business logic (.e.g link run to old algorithm GUIs)
> Let me know if it's ok for you

OK

luipir:
linked old GUI to each run button (depending if it run locally or remotely)
Screenshot from 2021-03-30 17-26-22

luipir:
not clear when to enable both  "Run" and "Run with default data"... the GUI is the same in both cases

LDMP/main_widget.py
```diff
@@ -54,10 +62,18 @@ def __init__(self, parent=None):
...
         self.treeView_datasets.setIndentation(0)
         self.treeView_datasets.verticalScrollBar().setSingleStep(10)
 
+        # instantiate calcluate callback container to reuse most of old code
+        # do this before setting setupAlgorithmsTree
+        self.dlg_calculate_LD = DlgCalculateLD()
+        self.dlg_calculate_TC = DlgCalculateTC()
+        self.dlg_calculate_Biomass = DlgCalculateRestBiomass()
+        self.dlg_calculate_Urban = DlgCalculateUrban()
+
```
ricardogsilva, line 76: This is very slow and introduces a noticeable delay in the UI from the moment the user presses the button to open the plugin dock until it is shown on screen. Please use a different strategy for loading these widgets without such a noticeable block of the main QGIS UI

LDMP/main_widget.py
```diff
@@ -121,56 +137,210 @@ def setupDatasets(self):
...
+            details=None,
+            parent=land_degradation_group,
+            run_callbacks={AlgorithmRunMode.Remotely: self.dlg_calculate_LD.btn_prod_clicked})
+        land_productivity_alg.run_mode = AlgorithmRunMode.Remotely
+        land_productivity_alg_details = AlgorithmDetails(
+            name=None,
+            name_details=None,
+            description=tr('TODO: Land productivity long description'),
+            parent=land_productivity_alg)
+        land_productivity_alg.setDetails(land_productivity_alg_details)
```
ricardogsilva, line 160: This API seems contrived and we ought to make it simpler. 

Do we really need `AlgorithmDetails` to be a class? And what is the meaning of it having `name`, `name_details`,  `parent`? Seems like this complexity is just for satisfying Qt's MVC stuff. IMO we can use something simpler to understand as an algorithm's configuration and then transform it to whatever Qt needs in order to show it in the tree.

LDMP/main_widget.py
```diff
@@ -121,56 +137,210 @@ def setupDatasets(self):
     def setupAlgorithmsTree(self):
```
ricardogsilva, line 137: This method is huge. Please break this up into smaller pieces so that it becomes more readable.

LDMP/main_widget.py
```diff
@@ -54,10 +62,18 @@ def __init__(self, parent=None):
...
 
+        # instantiate calcluate callback container to reuse most of old code
+        # do this before setting setupAlgorithmsTree
+        self.dlg_calculate_LD = DlgCalculateLD()
+        self.dlg_calculate_TC = DlgCalculateTC()
+        self.dlg_calculate_Biomass = DlgCalculateRestBiomass()
+        self.dlg_calculate_Urban = DlgCalculateUrban()
+
         self.setupAlgorithmsTree()
         self.setupDatasets()
```
ricardogsilva, line 78: Populating the list of datasets should be made async since it is a potentially long operation (it will involve scanning the user's hard disk and also contacting the remote GEE server). I realize we are not at that stage yet but nevertheless we can already start laying the ground for it.

Also, perhaps we can defer setting up the existing datasets to when the user actually visits the datasets tab instead of doing this work up front?

ricardogsilva (changes requested):
It is a commendable effort, but I believe it still requires further work. Some more comments, in addition to the ones inline:

- I could not actually run an algorithm yet, which is the whole point of this PR I guess. When pressing the Run button in the algorithm dialog, it doesn't show me any feedback on what is happening and eventually I need to force QGIS to quit.

- The run button on each algorithm in the tree should only be compound in case the algorithm can be run multiple ways. If it cannot, then it should be a normal button

luipir:
> It is a commendable effort, but I believe it still requires further work. Some more comments, in addition to the ones inline:
> 
> * I could not actually run an algorithm yet, which is the whole point of this PR I guess. When pressing the Run button in the algorithm dialog, it doesn't show me any feedback on what is happening and eventually I need to force QGIS to quit.
> * The run button on each algorithm in the tree should only be compound in case the algorithm can be run multiple ways. If it cannot, then it should be a normal button

luipir:
sorry closed by mistake

LDMP/main_widget.py
luipir: 1: Algorithm details can be more that a simple description... I would leave it as container
2: the original idea was to use details inside a widget (was the first really hugly GUI imlementation)... then simplified the MVC leaved organizing to the TreeView but required to add a class/model level.
The fact that remianed name and name_details is because the original base class had name and name_detaisl becase wasn0't expected to have details ans node... we can remove. But probably we are spending too much time in representation than running the algorithm.Imho a second core part woudl be to refactor and simplify this part... creating a good widget as node (more or less as in Dataasets) but time is short to focus completly in the estetics.

LDMP/main_widget.py
luipir: Yeah... I had to instantiate it before to use the click event wrapper of the interface. btw I can use a lambda so layer instantiate it.
Please open an issue about this

luipir:
> * I could not actually run an algorithm yet, which is the whole point of this PR I guess. When pressing the Run button in the algorithm dialog, it doesn't show me any feedback on what is happening and eventually I need to force QGIS to quit

??? strange! are you talking the run button of the algorithm in the dock or the run in the algorithm interface? e.g. the ord run algorithm interface.

LDMP/main_widget.py
luipir: ok, I'll do when I've the API interface to download algorithms/datasets

luipir:
> * I could not actually run an algorithm yet, which is the whole point of this PR I guess. When pressing the Run button in the algorithm dialog, it doesn't show me any feedback on what is happening and eventually I need to force QGIS to quit.

@ricardogsilva Checked and it works... there were only an error casting buffer ins settings. can you share more details about what happen to you?

LDMP/main_widget.py
luipir: I did lazy instantiation of the widget in the delegate (that is in charge of the run the callback) but I've some strange side effects... I'll give a look later. for now please leave it as it is and move the lazy instantiation in a separate issue.

ricardogsilva:
@luipir 

Here is a short video of me trying to run an algorithm after I had updated to the latest commit of this PR's branch. After I press the button to start executing the algorithm, the algorithm dialogue stays open and after a while I just get a message saying QGIS is not responding.

https://user-images.githubusercontent.com/732010/114578523-dcd4b900-9c74-11eb-8b14-f4c314180c41.mp4

luipir:
> Here is a short video of me trying to run an algorithm after I had updated to the latest commit of this PR's branch. After I press the button to start executing the algorithm, the algorithm dialogue stays open and after a while I just get a message saying QGIS is not responding.

GULP! should we have to do a interactive session to understand what's happening. Just for you info I'm running the plugin in master.
After push calculate, should start the progress bar of api request in the main window. are logged in or setup auth configuration? if not, should return an error I suppose.

ricardogsilva:
After we had our interactive debug session we could verify that the error I get about not being able to run any algorithm seems to be related to `QgsSettings`, and indeed out of scope for this PR - check #389 

As such I am accepting this PR.

PR #391, opened 2021-04-14
Cast to bool boolean key in the ini file. Fixes #389

### LINKED ISSUE #389: Invalid QgsSettings cause the plugin to not be able to run any algorithm
I have somehow managed to save some incorrect state into my `QgsSettings` and now the plugin is unable to run its algorithms. I am attaching a copy of the `.ini` file saved by QGIS in order to help debugging (had to put it inside a zip in order to trick github).

If I try to run a TE algorithm with these QSettings in place my QGIS just freezes and I eventually am forced to kill it.

When a new QGIS profile is used, one with also a new QgsSettings object, the algorithm execution error does not manifest - If I start a new QGIS profile, install TE into it and try to run an algorithm, execution fails in a wel behaved manner, showing me a pop up dialogue with the following message:

```
Choose an area of interest.
```

So it seems that the original `QgsSettings` have some error in the Trends.Earth region of interest.

QGIS settings file: [qgis-settings-ini.zip](https://github.com/ConservationInternational/trends.earth/files/6310150/qgis-settings-ini.zip)

luipir:
fixes #389

### DIFF
--- a/LDMP/settings.py
+++ b/LDMP/settings.py
@@ -273,7 +273,7 @@ class AreaWidget(QtWidgets.QWidget, Ui_WidgetSelectArea):
 
     def load_settings(self):
 
-        buffer_checked = self.settings.value("trends_earth/region_of_interest/buffer_checked", False, type=bool) == 'True'
+        buffer_checked = self.settings.value("trends_earth/region_of_interest/buffer_checked", False, type=bool)
         area_from_option = self.settings.value("trends_earth/region_of_interest/chosen_method", None)
 
         if area_from_option == 'country_region' or \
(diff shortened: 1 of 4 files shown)


LDMP/settings.py
```diff
@@ -273,7 +273,7 @@ def __init__(self, parent=None):
 
     def load_settings(self):
 
-        buffer_checked = self.settings.value("trends_earth/region_of_interest/buffer_checked", False) == 'True'
+        buffer_checked = self.settings.value("trends_earth/region_of_interest/buffer_checked", False, type=bool) == 'True'
```
ricardogsilva, line 276: This line is flawed, you are comparing against a string but at the same time you cast the output of `settings.value` to a `bool`. This test is always false. I suggest this instead:

```suggestion
        buffer_checked = self.settings.value("trends_earth/region_of_interest/buffer_checked", False, type=bool)
```

ricardogsilva (changes requested):
Nice work finding the root of the problem :smile_cat: 

Please fix the faulty check in the inline comment and I'll be :+1: to merge

LDMP/settings.py
luipir: good catch :)... I'll add a modification to a similar code in the same PR

luipir:
@ricardogsilva please review again... I removed all compare to 'True? all around the code (damn I've all the monitor flickering)

ricardogsilva (approved):
LGTM!

PR #392, opened 2021-04-15
User interface refactoring

luipir:
This PR has an extensive refactoring of both the UI and also of the internal architecture of the Trends.Earth plugin. Featured:

- Plugin now uses a dock widget as its main area
- Indicators are now presented by means of a tree
- Indicator execution method is now more clearly specified in the plugin UI
- Region of interest is only set in the settings dialogue
- Authentication with the remote Trends.Earth processing server leverages the QGIS auth framework
- Remote execution progress is monitored automatically (can be disabled in the settings)
- Remote results are downloaded automatically (can be disabled in the settings)
- All execution requests, both remote and local now go through the `job_manager`
- State is kept in-memory but is also synced with the local filesystem. Synchornization is done periodically

Most changes have already been presented and discussed with @azvoleff 

Here is a brief video demonstrating the overall plugin functionality:

https://user-images.githubusercontent.com/732010/122979516-8d56eb80-d38f-11eb-970a-37f04d672401.mp4


ricardogsilva:
@azvoleff 

We are continuing to work on this branch as we work through our implementation plan.

LDMP/utils.py
```diff
@@ -0,0 +1,85 @@
+import importlib
+import tempfile
+import typing
+from pathlib import Path
+
+import qgis.core
+from osgeo import gdal
+from openpyxl.drawing.image import Image
```
azvoleff, line 8: This throws an error on initial loading the plugin as `utils` is imported from at the beginning __init__.py, but `openpyxl` is potentially not available until later, after `ext-libs` is loaded. This can be fixed by moving `from openpyxl.drawing.image import Image` to be within the `maybe_add_image_to_sheet` function in `__init,py__`. Likely this only comes up on Windows or other systems where QGIS is using its own ython install that users haven't modified by adding any further dependencies (like openpyxl)

Samweli: @azvoleff I couldn't reproduce the error when installing the plugin on my window OS, could you give the steps to get the error and/or screenshot of the error, @ricardogsilva has opened the issue here https://github.com/ConservationInternational/trends.earth/issues/490, thanks.

Issue #409, opened 2021-05-03
Implement scanning of the local base dir in order to populate the datasets section


ricardogsilva:

luipir:
Probably we should avoid to update Jobs and Dataset asyncronously... or we should add a lock to avoid datasets created when the job store is updated.
A simplified solution is the unique refresh for both jobs and datasets (as it is now).
Do we really need a different refresh between jobs and datasets @ricardogsilva ?

ricardogsilva:
My two requirements regarding this are:

1. Avoid locking up the QGIS GUI while doing potentially expensive IO operations
2. Make it possible to control (and even disable) the frequency of checking the remote server while making sure that the frequency of checking the local base dir is never changed (nor disabled)

I will be OK with the actual technical solution, as long as it is able to meet the above requirements.

---

Thinking a bit about it, I guess we should also explicitly trigger a refresh of the local datasets (_i.e._ an additional check, outside of the normal periodic checks) list when:

- One of the local algorithms finishes execution. 
- The download of a dataset finishes

This shall help achieving a lower latency as to when a dataset appears on the list after being available. 

I don't really see a need for using locks in this case. The scan of the base dir shall be looking for JSON files and parsing their content in order to read the dataset's `status` (we don't need to scan for the actual geotiff files nor do we need to wonder if the download is finished or not). If the JSON file reports that the dataset is available, then we put that dataset in our list as being loadable. Otherwise, the dataset's status shall report if the dataset is still being generated or being downloaded, and in that case we put it on the list as not being available.

luipir:
> * One of the local algorithms finishes execution.

still have to check if all works with local process :/

luipir:
local process is done correctly now:
implementatuion in branch https://github.com/luipir/trends.earth/tree/local_calculation

BTW generate results e.g. local generated dataset have (yet another) json format => need to manage (yet another) exception to allow visualize them in Dataset list.

@ricardogsilva what would be the expected way to manage locally generated datasets? e.g. what to show in Dataset list? what funcion have to be available? Add to canvas? dataset details? 
About Dataset details I would warn @Samweli to take into account this other type of dataset in case have to be shown in Dataset Details interface

luipir:
I leaved it open becase we still have to manage get of datasets generated by local processing

luipir:
wating for merge of https://github.com/ConservationInternational/trends.earth/pull/430 to close the issue

luipir:
[...]

Issue #407, opened 2021-05-03
Implement dataset details dialog


ricardogsilva:
Each dataset shall have a dedicated dialog where all of its known details are shown to the user. This shall list:

- Details related to the dataset (name, creation date, etc)
- Details of the algorithm that produced the dataset (algorithm parameter values, etc)
- Notes written by the user upon configuring the algorithm

The dataset dialog shall also feature some buttons to perform dataset-related actions:

- Load dataset onto QGIS canvas
- Delete dataset
- Export dataset report (as discussed on #388)

The below mockup shows initial design for this dialog

![geo-ldn-dataset-detail-dialogue2](https://user-images.githubusercontent.com/732010/116892212-816c6a00-ac27-11eb-9b9a-d2c684e0df6d.png)

luipir:
@Samweli I warn you to take into account locally generated datasets in case have to be shown in Dataset Details interface.
a JSON file genrated for a local process can have this form (but not sure if it is the only schema generated for local generated dataset)

```
{
    "bands": [
        {
            "activated": true,
            "add_to_map": false,
            "metadata": {},
            "name": "SDG 15.3.1 Indicator",
            "no_data_value": -32768.0
        },
        {
            "activated": true,
            "add_to_map": false,
            "metadata": {},
            "name": "SDG 15.3.1 Productivity Indicator",
            "no_data_value": -32768.0
        }
    ],
    "file": "8c7637b2-77d0-44f6-a7b5-da015ec996b2.tif",
    "metadata": {
        "task_name": "gen_final1",
        "task_notes": ""
    }
}
```

Issue #373, opened 2021-02-25
Allow disabling automatic download of remotely-generated datasets


ricardogsilva:
By default datasets generated remotely are to be automatically downloaded when they are ready. This may be undesirable in some circumstances. As such. lets also provide an option in the settings to disable this behavior and make the user explicitly ask for a dataset to be downloaded.

This new setting shall be put under the **Advanced** section.

In addition, the dataset list shall display an additional `Download` button for datasets that are ready to be downloaded. Once the dataset is available locally (_i.e._ its state becomes `available`) the button shall no longer be shown. (In a future work item, when implementing the dataset details dialogue, we shall show a more permanent download button)

PR #412, opened 2021-05-05
Automatic/configurable datasets download. Fixes #376 #409

### LINKED ISSUE #376: Implement automatic download of remote datasets
Whenever a remote algorithm finished execution its outputs become ready for download. The default behavior for Trends.Earth shall be to proceed to download them automatically.

Foreseen workflow is:

- A refresh of the datasets list is initiated (either automatically or by request of the user)
- Datasets that are in a `being generated` state are checked to discover whether they are ready to be downloaded
- trends.earth proceeds to download datasets that are ready, updating their state accordingly

Downloading a dataset shall be made in a background task, in order to prevent locking up the QGIS main UI and allow the user to carry on with other work.

Additionally we shall allow this automatic download behavior to be disabled by the user. This shall be tracked in #373 and is not in scope for this issue.

luipir:
This PR add automatic download of datasets that are in FINISHED or SUCCESS state.
Autodownload can be disabled setting the boolean config parameter:

`trends_earth/advanced/dataset_auto_download`

that is not present is assumed True => auto download
After download download state is updated to the next Dataset refresh from base_data_directory

Downloaded dataset are not downloaded again if already downloaded BTW the original dataset json descriptor is not removed, just ignored is the equivalent downloaded json descriptor is already present.

This PR shoudl fix the following issue/features:
https://github.com/ConservationInternational/trends.earth/issues/376
https://github.com/ConservationInternational/trends.earth/issues/409

### DIFF
--- a/LDMP/models/datasets.py
+++ b/LDMP/models/datasets.py
@@ -395,6 +395,12 @@ class Datasets(QObject):
                     # if strored in datasetStore remove entry related to json_file Dataset
                     self.datasetsStore.pop(json_file, None)
 
+                    # remove also the dataset json that now is nomore necessary
+                    try:
+                        os.remove(json_file)
+                    except:
+                        pass
+
                     log(tr('Dataset {} already downloaded => skipped').format(json_file))
                     continue
 
--- a/LDMP/models/datasets.py
+++ b/LDMP/models/datasets.py
@@ -371,9 +371,15 @@ class Datasets(QObject):
                 #   b4b098f8-e562-4843-b6ed-586878410a01.json
                 # a downloaded dataset could be:
                 #   b4b098f8-e562-4843-b6ed-586878410a01_Productivity 1_0_3_gen_test19.json
-                basename = os.path.basename(json_file)
+                basename = os.path.splitext(os.path.basename(json_file))[0]
                 look_for = basename + "_"
-                if look_for in json_files:
+
+                found = next((j for j in json_files if look_for in j), None) # look if some key containing 'look_for' string
+                if found:
+                    # if strored in datasetStore remove entry related to json_file Dataset
+                    self.datasetsStore.pop(json_file, None)
+
+                    log(tr('Dataset {} already downloaded => skipped').format(json_file))
                     continue
 
                 # block useful to distinguish if dataset is a downloaded dataset or not
--- a/LDMP/models/datasets.py
+++ b/LDMP/models/datasets.py
@@ -222,6 +222,16 @@ class Dataset(DatasetBase):
     def fileName(self):
         return self.__fileName
 
+    @staticmethod
+    def datetimeRepr(dt: datetime) -> str:
+        return datetime.strftime(dt, '%Y/%m/%d (%H:%M)')
+
+    @staticmethod
+    def toDatetime(dt: str, fmt: str = None) -> datetime:
+        if fmt is None:
+            return datetime.fromisoformat(dt)
+        return datetime.strptime(dt, fmt)
+
     def download(self, datasets_refresh: bool = True) -> None:
         """Download Dataset related to a specified Job in a programmatically defined filename and folder.
         Because a new dataset JSON descriptor is created => Datasets sync
--- a/LDMP/models/datasets.py
+++ b/LDMP/models/datasets.py
@@ -262,7 +262,7 @@ class Dataset(DatasetBase):
         log(u"Processing job {}.".format(job.runId))
         result_type = job.results.get('type')
         if result_type == 'CloudResults':
-            download_cloud_results(job.raw, f, self.tr)
+            download_cloud_results(job.raw, f, self.tr, add_to_map=False)
             self.downloaded.emit(f)
             Datasets().sync()
         elif result_type == 'TimeSeriesTable':
--- a/LDMP/models/datasets.py
+++ b/LDMP/models/datasets.py
@@ -14,7 +14,7 @@
 __author__ = 'Luigi Pirelli / Kartoza'
 __date__ = '2021-03-23'
 
-from typing import Optional, Union, List, Dict
+from typing import Optional, Union, List, Dict, Tuple
 from datetime import datetime
 from enum import Enum
 import abc
@@ -24,7 +24,7 @@ from collections import OrderedDict
 
 from qgis.PyQt.QtCore import QSettings, pyqtSignal, QObject
 from qgis.core import QgsLogger
-from LDMP.jobs import Job, JobSchema
+from LDMP.jobs import Job, JobSchema, Jobs, download_cloud_results, download_timeseries
 from LDMP.calculate import get_script_group
 from LDMP import log, singleton, tr, json_serial
 
@@ -91,13 +91,20 @@ class DatasetBase(abc.ABC, QObject, metaclass=FinalMeta):
 class Dataset(DatasetBase):
 
     dumped = pyqtSignal(str)
+    class Origin(Enum):
+        job = 0,
+        dataset = 1,
+        downloaded_dataset = 2,
+        not_applicable = 3
 
     def __init__(self,
             job: Optional[Job] = None,
-            dataset: Optional[Dict] = None
+            dataset: Optional[Dict] = None,
+            filename: Optional[str] = None
         ) -> None:
         super().__init__()
 
+        self.__origin = Dataset.Origin.not_applicable
         if job:
             job_schema = JobSchema()
             self.job = job_schema.dump(job)
@@ -109,18 +116,46 @@ class Dataset(DatasetBase):
             self.name = job.taskName
             self.creation_date = job.startDate
             self.run_id = job.runId
-        elif dataset:
-            self.job = ''
 
-            self.status = dataset.get('status')
-            # self.status: DatasetStatus = getStatusEnum(job.status)
-            self.progress = dataset.get('progress')
-            self.source = dataset.get('source')
-            self.name = dataset.get('name')
-            self.creation_date = dataset.get('creation_date')
-            self.run_id = dataset.get('run_id')
+            self.__origin = Dataset.Origin.job
+            self.__fileName = None
+
+        elif dataset:
+            # check if refer to a Dataset or Downloaded dataset
+            if 'bands' in dataset:
+                self.bands = dataset.get('bands')
+                self.file = dataset.get('file')
+                self.metadata = dataset.get('metadata')
+
+                # set common dataset data get from medatata (e.g. original Job)
+                self.status = self.metadata.get('status')
+                self.progress = self.metadata.get('progress')
+                self.source = self.metadata.get('script_name')
+                self.name = self.metadata.get('task_name')
+                self.creation_date = self.metadata.get('start_date')
+                self.run_id = self.metadata.get('id')
+
+                self.__origin = Dataset.Origin.downloaded_dataset
+                self.__fileName = filename
+
+            else:
+                self.job = ''
+
+                self.status = dataset.get('status')
+                # self.status: DatasetStatus = getStatusEnum(job.status)
+                self.progress = dataset.get('progress')
+                self.source = dataset.get('source')
+                self.name = dataset.get('name')
+                self.creation_date = dataset.get('creation_date')
+                self.run_id = dataset.get('run_id')
+
+                self.__origin = Dataset.Origin.dataset
+                self.__fileName = filename
         else:
             raise Exception('Nor input Job or dataset dictionary are set to build a Dataset instance')
+
+    def origin(self) -> 'Dataset.Origin':
+        return self.__origin
 
     def columnCount(self) -> int:
         return 1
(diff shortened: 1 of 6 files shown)


luipir:
rebased including the new @Samweli feature merged in decision-tree branch

LDMP/models/datasets.py
```diff
@@ -178,8 +215,80 @@ def dump(self) -> str:
...
+        base_dir = os.path.dirname(dataset_filename)
+        f = None
+        if job.results.get('type') != 'TimeSeriesTable':
+            # create result filename basing on:
+            # run_id + scriptName + task_name (if any)
+            fileBaseName = None
+            if job.taskName:
+                fileBaseName = u'{}_{}_{}'.format(job.runId, job.scriptName, job.taskName)
+            else:
+                fileBaseName = u'{}_{}_{}'.format(job.runId, job.scriptName)
```
ricardogsilva, line 271: This line is flawed as there are only two arguments to the `format()` function but the string specifies three placeholders. In my case, since I did not assign a name to the sample job I created, it makes the plugin crash

```suggestion
                fileBaseName = u'{}_{}'.format(job.runId, job.scriptName)
```

ricardogsilva (dismissed):
It is great to see our design starting to materialize, nice work!

I have a couple of requests, in addition to the bug which I mention inline:

- If the download is set to be automatic (which is the default behaviour) then the download button should not be shown at all. It should only be shown when the user has disabled automatic downloads (as per #373). Even then, the download button should be disabled until we know that the job has finished running

- When downloading finishes, the download button still remains visible until the next time the remote server is checked. It should be hidden immediately, since we even show a message to the user to let him know the file has been downloaded. Otherwise it becomes confusing for the user, since it seems that the button should be clicked.

- When download finishes, I see two JSON files are present in the filesystem, under `{base-dir}/outputs/SDG15.3.1/{creation-date}`:

  - job file
  - downloaded dataset file

  Seems to me that the job file is no longer relevant, after the dataset has been downloaded. The dataset JSON file seems to contain the relevant details about the dataset and also about the job that originated it. Is this correct? If so, please remove the job file

Additional notes, that we ought to address, in separate issues:

- Since we are not able to get a ganular enough progress report to show the execution percentage when the dataset is being generated (this was confirmed by @azvoleff  in the previous meeting) it would be best to show a message like `PROCESSING` instead of the progress bar. Or maybe show a progress bar with the infinite loading status to denote it is doing something but we can't be more specific about the progress?

- This is a minor nitpick but the widgets move slightly when the download icon is removed. It would be nicer if they didn't.

LDMP/models/datasets.py
```diff
@@ -91,13 +91,21 @@ def child(self, row) -> None:
 class Dataset(DatasetBase):
 
     dumped = pyqtSignal(str)
+    downloaded = pyqtSignal(str)
+    class Origin(Enum):
```
Samweli, line 95: I would prefer to define the Origin enum outside the `Dataset` class

LDMP/models/datasets.py
```diff
@@ -178,8 +215,80 @@ def dump(self) -> str:
...
+            f = os.path.join(base_dir, fileBaseName)
+            log(u"Downloading results to {} with basename {}".format(os.path.dirname(f), os.path.basename(f)))
+
+        log(u"Processing job {}.".format(job.runId))
+        result_type = job.results.get('type')
+        if result_type == 'CloudResults':
+            download_cloud_results(job.raw, f, self.tr, add_to_map=False)
+            self.downloaded.emit(f)
+            if datasets_refresh:
+                Datasets().sync()
```
Samweli, line 283: shouldn't we make this a static method?

Samweli (commented):
I think we need to choose one naming convention that should be use, I see currently there is a mix of camel case with snake case, also we need to fix typos in documentation as they can mislead the intended interpretation.

luipir:
> * Even then, the download button should be disabled until we know that the job has finished running

no button is available until finished

luipir:
> * When downloading finishes, the download button still remains visible until the next time the remote server is checked. It should be hidden immediately, since we even show a message to the user to let him know the file has been downloaded. Otherwise it becomes confusing for the user, since it seems that the button should be clicked.

just because the datasets item is refreshed at the end of refresh rate... I can disable or hide the download if set auto download

luipir:
> * f so, please remove the job file

ok, I leaved it just in case and traceability

luipir:
> * Or maybe show a progress bar with the infinite loading status

good idea

LDMP/models/datasets.py
luipir: rationale?

LDMP/models/datasets.py
luipir: could be static because it's a singleton working on class variable... btw I would avoid to access it without a single instance on which to work. e.g I would leave it as singleton to force user use singleton constructor and explicitly dige the idea they are working on a single instance

ricardogsilva:
> > * Even then, the download button should be disabled until we know that the job has finished running
> 
> no button is available until finished

What I meant was, when implementing the changes for #373 it would be good if the download button then becomes visible nut disabled while the dataset is being prepared (so that the user is able to discover that at some point he will be able to download the dataset manually). Then, when available, the download button becomes enabled, so that the user is able to download it. Finally, when the user has already downloaded, the button is hidden.

But again, this shall be only for when the user has disabled automatic downloads.

luipir:
> becomes visible but disabled while the dataset is being prepared (so that the user is able to discover that at some point he will be able to download the dataset manually)

now it's clear

luipir:
> > > * Even then, the download button should be disabled until we know that the job has finished running
> > 
> > 
> > no button is available until finished
> 
> What I meant was, when implementing the changes for #373 it would be good if the download button then becomes visible but disabled while the dataset is being prepared (so that the user is able to discover that at some point he will be able to download the dataset manually). Then, when available, the download button becomes enabled, so that the user is able to download it. Finally, when the user has already downloaded, the button is hidden.
> 
> But again, this shall be only for when the user has disabled automatic downloads.

should be done now

luipir:
> * This is a minor nitpick but the widgets move slightly when the download icon is removed. It would be nicer if they didn't.

should be done now

luipir:
> * Since we are not able to get a ganular enough progress report to show the execution percentage when the dataset is being generated (this was confirmed by @azvoleff  in the previous meeting) it would be best to show a message like `PROCESSING` instead of the progress bar. Or maybe show a progress bar with the infinite loading status to denote it is doing something but we can't be more specific about the progress?
> *

set "Processing..." and not dynamic progressbar. the tech reason is that any itme in the view is a statically rendered widget that become a working widget (editor) when over it with the mouse. In this way is hard to have widget repaint.

luipir:
> I think we need to choose one naming convention that should be use, I see currently there is a mix of camel case with snake case, also we need to fix typos in documentation as they can mislead the intended interpretation.

I agree... I usually use camel case, but part of the original code is snake case => what to follow?

luipir:
> * f so, please remove the job file

should be done now

ricardogsilva (changes requested):
Some things still need some additional modification:

1. When automatic download is enabled, the download button should never be visible

2. Periodic polling of the remote server does not seem to perform the same actions as manually pressing the refresh button. If I let QGIS automatically poll the remote GEE server then it never becomes aware that the processing is ready and it never downloads the results. I found that I have to explicitly press the 'refresh' button once in order to have QGIS become aware of the availability of the results. Even then, I still need to wait some time before the file is actually downloaded and for the datasets list to be updated.

Here is a small video I made depicting this problem with the automatic polling


https://user-images.githubusercontent.com/732010/117475456-f99e9c80-af53-11eb-87c9-11b4620c25ce.mp4

luipir:
> 2\. Periodic polling of the remote server does not seem to perform the same actions as manually pressing the refresh button. If I let QGIS automatically poll the remote GEE server then it never becomes aware that the processing is ready and it never downloads the results. I found that I have to explicitly press the 'refresh' button once in order to have QGIS become aware of the availability of the results. Even then, I still need to wait some time before the file is actually downloaded and for the datasets list to be updated.

Yes... I realised I forgot to trigger the periodic singleShot....

luipir:
I've rebased after merge of icen PR by @Samweli 
Now fixing the features not yet completed @ricardogsilva

LDMP/models/datasets.py
Samweli: We should have one place where all the enums are defined, eg in the dataset model class.

luipir: I would agree... but the other enum (DatasetStatus) will be removed because unused (as described in the TODO) ... just leaved in case we change idea.
I put the enum in the class just to leave clear the contect of the origin (e.g. is Dataset), otherwise I should name it DatasetOrigin, and IMHO it's more contextual aware having it accessible as class variable of the Dataset class. 
BTW I'm ok also to move outside if this is important for you.
Let mw knoe your decision.... it's up to you.

luipir:
> 1. When automatic download is enabled, the download button should never be visible
> 2. Periodic polling of the remote server does not seem to perform...

done

Samweli:
> 2\. Periodic polling of the remote server does not seem to perform the same actions as manually pressing the refresh button. If I let QGIS automatically poll the remote GEE server then it never becomes aware that the processing is ready and it never downloads the results. I found that I have to explicitly press the 'refresh' button once in order to have QGIS become aware of the availability of the results. Even then, I still need to wait some time before the file is actually downloaded and for the datasets list to be updated.

@luipir I have pulled the latest changes but I still have this issue.

luipir:
> I have pulled the latest changes but I still have this issue.

checking if I missed something

ricardogsilva (changes requested):
I still get some funky behavior with the latest changes:

1. I ask for an algorithm to run
1. Algorithm starts running. I can see that automatic polling of the remote server is running by looking at the QGIS console
1. Sometime later, I receive an email informing me that algorithm execution has finished. Meanwhile QGIS is still not aware of this, which I guess is understandable. I would expect that by the next time Trends.Earth polls the remote server, it also becomes aware that execution is done
1. I can see on the QGIS console that TE does a new poll of the remote server AFTER I have received this email, so I would think the remote server now reports to T.E. that execution is finished. Regardless, T.E. still does not mention that execution is finished
1. After waiting one more minute, T.E does an additional poll of the remote server and now it seems to finally be aware that the remote server has finished execution and proceeds to automatically download the result - this is more than two minutes after I got the email saying that execution is finished
1. I see that automatic download of the result is working now, which is good
1. After having the result downloaded locally, I see that the job file is still present on the base dir. I need to wait another minute, at the time when T.E does another update to get the job file to be deleted. I think the job file should be immediately deleted after the download is done, because we will have the result JSON file now

luipir:
> 3\. I would expect that by the next time Trends.Earth polls the remote server, it also becomes aware that execution is done

me too... but I  update basing on what is the answer from server side (1' is the default polling time). T.E is passive in this case.

Samweli:
> @Samweli are you sure you are using the latest modification? seems it works for @ricardogsilva

@luipir I have pulled the latest changes and it works when I submit new downloads, but the old ones have no progress anymore.

luipir:
> 7\. After having the result downloaded locally, I see that the job file is still present on the base dir. I need to wait another minute, at the time when T.E does another update to get the job file to be deleted. I think the job file should be immediately deleted after the download is done, because we will have the result JSON file now

with latest commit https://github.com/ConservationInternational/trends.earth/pull/412/commits/1e6097ab462ecdadf485297de8033bb077e0d7aa should remove immediatly all json that has been downloaded. Effectively I used the wrong update workflow. just updating the model but not synch datasets with that available in the folder => good catch @ricardogsilva

luipir:
> but the old ones have no progress anymore

@Samweli has been downloaded automatically => no more progress bar? Or I do not understand what you mean? can you share snapshot?

luipir:
> @luipir The download doesn't continue for old datasets and the progress bar is not removed, see ...

first step is to build Datasets form Jobs and Jobs are:
1) or downloaded from server
2) get from cache that is stored in qgis.ini... cache is updated when jobs are downloaded from server.

btw I'll check if there is a cleaning action of jobs from base data directory... I suppose there are old jobs remained when there where no possibility to remove them.
I'll check later... now I've to have a break for bureaucracy matters. In the meantime, please check in base_data_directry/Jobs if there are Jobs that are not listed in cache or available from server

luipir:
@Samweli probably it's time to implement it. https://github.com/luipir/trends.earth/blob/datasets_download_rebased/LDMP/jobs.py#L246
I leaved it becaseu I supposed was not enough important. Case as that you faces is due to the developing situation... shouldn't be a normal case. BTW I'll implement cleaning  of jobs if not present in current list of downloaded jobs from server.
Possible side effect! a Job and related Dataset disappear if not downloaded, btw if not downloaded after 15days probably was not so interesting.

Samweli:
> @Samweli I implemented cleaning of Jobs outside the cache here [luipir@daa8493](https://github.com/luipir/trends.earth/commit/daa84931ea618d80e92f0028749daa165feaf25f)
> I can't cherry-pick it because need to move traverse method in **init** and this is done in the following PR related to the delete of datasets.
> My proposal is to move to merge this PR, then that that come from https://github.com/luipir/trends.earth/tree/dataset_delete (#416) and at the end that that manage dataset_add where I added the fix you asked

@luipir yes we should move forward with this as I also need to use the changes in those two other PRs in order to complete PR for https://github.com/ConservationInternational/trends.earth/issues/407

luipir:
> @luipir yes we should move forward with this as I also need to use the changes in those two other PRs in order to complete PR for #407

I agree... wait for review or thumb up by @ricardogsilva. If green light, please @Samweli provide to merge if there are no other modifications

ricardogsilva (approved):
Seems to be working OK now, so I'm +1 for merging.

Just take care of fixing the conflicts reported by github, as it seems to not allow auto merge with the current state

ricardogsilva:
@Samweli I see this still needs you review and approval?

Samweli:
@ricardogsilva I have already reviewed it. @luipir can you update the PR with the latest changes from the decision-trees branch. Then we can merge it.

luipir:
> @ricardogsilva I have already reviewed it. @luipir can you update the PR with the latest changes from the decision-trees branch. Then we can merge it.

doing now

ricardogsilva:
@Samweli if you already reviewed, then please use the github review feature and mark the PR as accepted

luipir:
@Samweli rebased over latest decision-tree updates

luipir:
@samweli after accept review, please merge

Samweli (approved):

Issue #402, opened 2021-05-03
Allow loading a dataset onto QGIS canvas


ricardogsilva:
Each available dataset in the dataset list shows a button to load it onto QGIS canvas. Hook up this loading functionality by reusing the code, which already exists.

### NOTE

Datasets that are not available locally shall have the load button disabled

luipir:
dev in branch https://github.com/luipir/trends.earth/tree/dataset_add

PR #418, opened 2021-05-11
Dataset add

luipir:
Add a dataset to canvas. fixes https://github.com/ConservationInternational/trends.earth/issues/402

### DIFF
--- a/LDMP/jobs.py
+++ b/LDMP/jobs.py
@@ -420,6 +420,7 @@ def download_result(url, out_file, job, expected_etag):
 def download_cloud_results(job, f, tr, add_to_map=True):
     results = job['results']
     json_file = f + '.json'
+    downloaded = False
     if len(results['urls']) > 1:
         # Save a VRT if there are multiple files for this download
         urls = results['urls'] 
@@ -436,6 +437,7 @@ def download_cloud_results(job, f, tr, add_to_map=True):
                                    binascii.hexlify(base64.b64decode(urls[n]['md5Hash'])).decode())
             if not resp:
                 return
+            downloaded = True
         # Make a VRT mosaicing the tiles so they can be treated as one file 
         # during further processing
         out_file = f + '.vrt'
@@ -456,6 +458,7 @@ def download_cloud_results(job, f, tr, add_to_map=True):
                                 binascii.hexlify(base64.b64decode(url['md5Hash'])).decode())
             if not resp:
                 return
+            downloaded = True
 
     create_gee_json_metadata(json_file, job, out_file)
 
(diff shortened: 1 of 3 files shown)


luipir:
@ricardogsilva @Samweli this PR is ready to review

ricardogsilva (changes requested):
Seems to work OK.

However, I think there is no need to show a progress message to the user in this case. It becomes too noisy. Additionally, the message show says something like `Downloaded: ...`, which is misleading, since nothing is being downloaded. Please remove showing a progress message in this case.

Samweli (commented):
> However, I think there is no need to show a progress message to the user in this case. It becomes too noisy. Additionally, the message show says something like `Downloaded: ...`, which is misleading, since nothing is being downloaded. Please remove showing a progress message in this case.

I agree with @ricardogsilva, the right progress message should be about the dataset loading progress.

luipir:
I'll check. teoretically shouldn't download anything in case already downloaded or at least is this the logic of the original code.
I decided to hook the original code allow a more simple integration and a seamless passing from old code to the new one. e.g. better use the same method as before to allow old code continue working (just in case there are user that want to use the old interface). Another advantage is that the old code is well integrated with all other funcions => this reduce complexity in the refactoring.
I struggled to decide what to do... and at the end I decided to hook mainly because the overall complexity of the system is high (respect the time to develop). Another solution could be to extract part of the logic of add_layer avoind to download. But as far as I undestand the code, download was triggered in case not downloaded... I'll check!

luipir:
no sorry... I did a mistake. I use download() that is already an extract of original code with some modification about where to save staffs.
The last part of this code is the old logic using download_cloud_results or download_timeseries and there there is the downlod triggering for some reason.

luipir:
@ricardogsilva @Samweli are you sure is triggring download? because is using the download method  the log messages can be misleading. but nothig should be downloaded. the progress barr can be related to the polling to remote server.
I tested and no connection worker is reached during add later if already downloaded

luipir:
you can check for example set in qgis.ini setings
[trends_earth]
advanced\jobs_refresh_polling_time=0

in this way remote connection is disabled (after the next connection).

Samweli:
> @ricardogsilva @Samweli are you sure is triggring download?

@luipir it is not triggering download but it is showing a 'Downloaded..' message in the progress bar, I think it because of this line https://github.com/ConservationInternational/trends.earth/blob/2a147a5198f41161d4f39c63ca7c690aad2d9b96/LDMP/jobs.py#L469

luipir:
> > @ricardogsilva @Samweli are you sure is triggring download?
> 
> @luipir it is not triggering download but it is showing a 'Downloaded..' message in the progress bar, I think it because of this line
> 
> https://github.com/ConservationInternational/trends.earth/blob/2a147a5198f41161d4f39c63ca7c690aad2d9b96/LDMP/jobs.py#L469

exactly... I can move the message right after download to avoid misleading

luipir:
> @luipir it is not triggering download but it is showing a 'Downloaded..' message in the progress bar, I think it because of this line

modified sending message only if real download

ricardogsilva:
@luipir is this ready for another review?

luipir:
yes... sorry I forgot to notify via github

ricardogsilva (approved):
LGTM!

luipir:
so wait for @Samweli review and his merge

PR #421, opened 2021-05-12
fix Nonetype value passing error

Samweli:
This PR fixes an error that I kept getting which is related to progress update on an already downloaded Dataset. 
see
image

### DIFF
--- a/LDMP/models/datasets_delegate.py
+++ b/LDMP/models/datasets_delegate.py
@@ -152,20 +152,23 @@ class DatasetEditorWidget(QWidget, Ui_WidgetDatasetItem):
         self.labelRunId.setText(str(self.dataset.run_id)) # it is UUID
 
         # show progress bar or download button depending on status
-        self.progressBar.setValue( self.dataset.progress )
+        if self.dataset.progress is not None:
+            self.progressBar.setValue( self.dataset.progress )
         self.pushButtonStatus.hide()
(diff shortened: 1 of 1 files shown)


luipir:
@Samweli I suppose you had this issue because you have some local processed data. Isn't it? 

BTW better a fix that do some decision about values to show instead than do nothing.
In case of local processing we do not have status, progres nor source => all this data will be managed by the Datasets class when loading a JSON from local process.

If it is not a JOSN from local process => please give me info how did you generate it.

I would close this PR in favour of a more general management.

Samweli:
> @Samweli I suppose you had this issue because you have some local processed data. Isn't it?
> 
> BTW better a fix that do some decision about values to show instead than do nothing.

@luipir yes, I will be happy if we come with fix for it sooner :smile: , as the error keeps showing up to me while testing functionalities, and some times QGIS crashes!

luipir:
> error keeps showing up

I need to know in what case and with what JSOn you are facing this error.
BTW crash shouldn't be related with this issue

Samweli:
@luipir using the latest changes the error does not occur at all, I ll try to reproduce it, if not successful will close this.

Samweli:
@luipir @ricardogsilva closing this, the error doesn't occur anymore when using `decision-trees` branch latest changes,

Issue #400, opened 2021-05-03
Implement filtering the list of existing datasets


ricardogsilva:
The dataset list currently has a text box that shall be used to filter existing datasets. Filtering shall be done on a dataset's **name**.

Ideally this filtering should be done as the user is typing, _i.e._ without having to include another button on the UI to enable the filter string.

This may be implemented with [QSortFilterProxyModel](https://doc.qt.io/qt-5/qsortfilterproxymodel.html)

luipir:
closed via https://github.com/ConservationInternational/trends.earth/commit/8025f095b630dd74bf50a31816dffc1002524d7f

PR #423, opened 2021-05-12
Datasets filtering

Samweli:
Fixes https://github.com/ConservationInternational/trends.earth/issues/400

Implements Datasets filtering by adding a custom sort and filtering proxy model, a subclass of  [QSortFilterProxyModel](https://doc.qt.io/qt-5/qsortfilterproxymodel.html), the subclass handles a different filtering mechanism compared to its superclass as in our case we need to filter data from Dataset class.

I have updated the model item access in the `DatasetItemDelegate` to use `Qt.DisplayRole` instead of `Qt.ItemDataRole` because the [QSortFilterProxyModel](https://doc.qt.io/qt-5/qsortfilterproxymodel.html) only support the former.

The filtering is case insensitive.

screenshot showing Datasets filtering in action
filtering_datasets

### DIFF
--- a/LDMP/models/datasets.py
+++ b/LDMP/models/datasets.py
@@ -341,9 +341,6 @@ class Datasets(QObject):
     def columnName(self, column: int) -> Optional[str]:
         return None
 
-    def items(self):
-        return self.datasetsStore.items()
-
     def sync(self):
         """Method to sync with Datasets available in "trends_earth/advanced/base_data_directory".
 
--- a/LDMP/models/datasets.py
+++ b/LDMP/models/datasets.py
@@ -341,6 +341,9 @@ class Datasets(QObject):
     def columnName(self, column: int) -> Optional[str]:
         return None
 
+    def items(self):
+        return self.datasetsStore.items()
+
     def sync(self):
         """Method to sync with Datasets available in "trends_earth/advanced/base_data_directory".
 
--- a/LDMP/models/datasets_delegate.py
+++ b/LDMP/models/datasets_delegate.py
@@ -80,13 +80,13 @@ class DatasetItemDelegate(QStyledItemDelegate):
             return
 
         # activate editor
-        item = model.data(index, Qt.ItemDataRole)
+        item = model.data(index, Qt.DisplayRole)
         self.parent.openPersistentEditor(self.enteredCell)
 
     def paint(self, painter: QPainter, option: QStyleOptionViewItem, index: QModelIndex):
         # get item and manipulate painter basing on idetm data
         model = index.model()
-        item = model.data(index, Qt.ItemDataRole)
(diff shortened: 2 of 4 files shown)


luipir:
great

LDMP/models/datasets_delegate.py
```diff
@@ -80,13 +80,13 @@ def manageEditing(self, index: QModelIndex):
             return
 
         # activate editor
-        item = model.data(index, Qt.ItemDataRole)
+        item = model.data(index, Qt.DisplayRole)
```
luipir, line 83: I would prefer to separete Display and Item role... btw because, in this case, I used ItemDataRole just to create editor (that at the end is a kind of visualization). I accept this modification.

LDMP/models/datasets_model.py
```diff
@@ -56,7 +58,7 @@ def data(self, index: QModelIndex = QModelIndex(), role: Qt.ItemDataRole = Qt.Di
         item = index.internalPointer()
         if role == Qt.DisplayRole:
             if index.column() == 0:
-                return list(self.rootItem.items())[index.row()].name
+                return list(self.rootItem.items())[index.row()][1]
```
luipir, line 61: Because rootItem is a Datasets e.g.s a dict with key=json and value=Dataset and Dataset is stored as Item (see Datasets.child(...) method
I would simplify:
```suggestion
                return item
```
plus... I would remove line 60 because now unuseful

LDMP/models/datasets.py
```diff
@@ -341,6 +341,9 @@ def child(self, row: int) -> Optional['Dataset']:
     def columnName(self, column: int) -> Optional[str]:
         return None
 
+    def items(self):
```
luipir, line 344: method not necessary if accept modification in data(...) method

luipir (changes requested):
I suggest some modification to simplify the code... and please clarify me, whay do you only implemented filter by name and not also filter by start_date?

ricardogsilva:
> why do you only implemented filter by name and not also filter by start_date?

The original issue only mentions filtering datasets by name. IMO filtering by `start_date`, while interesting, is not a priority.

In case you are wondering, the combobox that is present on the UI next to the search text edit is to be used for sorting the list, and in there I think it does make sense to allow sorting by either name or date.

ricardogsilva (approved):
Functionality-wise it looks good to me.

Did not dive into the actual implementation. I leave that for @luipir

LDMP/models/datasets_delegate.py
Samweli: Thanks @luipir, the sort proxy model couldn't work with `Qt.ItemDataRole` that why I changed it.

luipir (approved):

Issue #401, opened 2021-05-03
Allow sorting the list of existing datasets according to various criteria


ricardogsilva:
The datasets section features a combo box for sorting but it currently does not work. Implement the sorting behavior. Make it possible to sort the datasets based on:

- name
- creation date
- originating algorithm
- execution status

Make it possible to also do a reverse sort. Introduce an additional checkbox in the UI in order to allow a user to specify this.

Ideally the sorting shall be done as soon as the user chooses one of the sort fields, _i.e._ without having to add an additional button to the UI to activate the sorting

This may be implemented with [QSortFilterProxyModel](https://doc.qt.io/qt-5/qsortfilterproxymodel.html)

PR #425, opened 2021-05-13
Datasets sorting

Samweli:
Fixes https://github.com/ConservationInternational/trends.earth/issues/401
Contains work from https://github.com/ConservationInternational/trends.earth/pull/423

Adds sorting functionality on the existing Datasets sort and filter proxy model, added a new Reverse checkbox that when checked the Datasets will be sorted in descending order. By default the sort in by creation date

I have added a progress message bar to notify user when Datasets sorting, this is useful when sorting take a bit long.
When datasets are refreshed and newer datasets have been added to the source model, they are automatically sorted by Date.

[**Updated**] Screenshot of sorting datasets example
updated_sorting

### DIFF
--- a/LDMP/models/datasets.py
+++ b/LDMP/models/datasets.py
@@ -597,7 +597,7 @@ class Datasets(QObject):
     def __less_than(self, left_dataset, right_dataset, field: SortField):
         if field == SortField.NAME:
             if left_dataset.name != right_dataset.name:
-                return left_dataset.name <= right_dataset.name
+                return left_dataset.name < right_dataset.name
         elif field == SortField.DATE:
             return self.__compare_dates(
                 left_dataset.creation_date,
@@ -605,10 +605,10 @@ class Datasets(QObject):
             )
         elif field == SortField.ALGORITHM:
             if left_dataset.source != right_dataset.source:
-                return left_dataset.source <= right_dataset.source
+                return left_dataset.source < right_dataset.source
         elif field == SortField.STATUS:
             if left_dataset.status != right_dataset.status:
-                return left_dataset.status <= right_dataset.status
+                return left_dataset.status < right_dataset.status
 
         return self.__compare_dates(
             left_dataset.creation_date,
--- a/LDMP/models/datasets.py
+++ b/LDMP/models/datasets.py
@@ -48,6 +48,13 @@ class DatasetStatus(Enum):
 DatasetStatusStrings = [e.name for e in DatasetStatus]
 
 
+class SortField(Enum):
+    NAME = 'name'
+    DATE = 'date'
+    ALGORITHM = 'algorithm'
+    STATUS = 'status'
+
+
 def getStatusEnum(status: str) -> DatasetStatus:
     """TODO: remove as commented above"""
     # get APIResponseSchema and remap into DatasetStatus
@@ -547,6 +554,46 @@ class Datasets(QObject):
         """Return Dataset and related descriptor asociated file."""
         datasets = [(k, d) for k,d in self.datasetsStore.items() if str(d.run_id) == id]
         return datasets[0] if len(datasets) else None
+
+    def sort(self, column, order, field: SortField):
+        self.datasetsStore = OrderedDict(self.__merge_sort(list(self.datasetsStore.items()), field))
+
+    def __merge_sort(self, items: List, field):
+        if len(items) <= 1:
+            return items
+        mid = int((len(items) / 2))
+
+        left = self.__merge_sort(items[:mid], field)
+        right = self.__merge_sort(items[mid:], field)
+        return self.__merge(left, right, field)
+
+    def __merge(self, left, right, field):
+        sorted_dict = []
+        i = j = 0
+
+        while i < len(left) and j < len(right):
+            if self.__less_than(left[i][1], right[j][1], field):
+                sorted_dict.append(left[i])
+                i += 1
+            else:
+                sorted_dict.append(right[j])
+                j += 1
+        sorted_dict.extend(left[i:])
+        sorted_dict.extend(right[j:])
+        return sorted_dict
+
+    def __less_than(self, left_dataset, right_dataset, field: SortField):
+        if field == SortField.NAME:
+            return left_dataset.name < right_dataset.name
+        elif field == SortField.DATE and \
+                isinstance(left_dataset.creation_date, datetime) \
+                and isinstance(right_dataset.creation_date, datetime):
+            return left_dataset.creation_date < right_dataset.creation_date
+        elif field == SortField.ALGORITHM:
+            return left_dataset.source < right_dataset.source
+        elif field == SortField.STATUS:
+            return left_dataset.status < right_dataset.status
+        return False
 
 
 class DatasetSchema(Schema):
--- a/LDMP/models/datasets.py
+++ b/LDMP/models/datasets.py
@@ -26,7 +26,7 @@ import glob
 import shutil
 
 from qgis.utils import iface
-from qgis.PyQt.QtCore import QSettings, pyqtSignal, QObject
+from qgis.PyQt.QtCore import QSettings, pyqtSignal, QObject, Qt
 from qgis.PyQt.QtWidgets import QMessageBox
 from qgis.core import QgsLogger
 from LDMP.jobs import Job, JobSchema, Jobs, download_cloud_results, download_timeseries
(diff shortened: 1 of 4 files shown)


ricardogsilva:
Before I do a review, I find it weird that sorting a list would require a progress bar. This should be a fast operation.

Please elaborate on why this is happening

Samweli:
> Before I do a review, I find it weird that sorting a list would require a progress bar. This should be a fast operation.

@ricardogsilva have you tested it and it is slow?

Samweli:
Rebased from decision-trees branch

luipir:
I can't see any performance issue in the code... really strange.
Can you check it there is some extra trigger of item rendering (method paint in the delegate) that can slow down sorting?

ricardogsilva (changes requested):
Sorting the list must be a fast operation - It is just a sort of something that is already in memory.

The current proposal takes too long and I'm not even sure it works. For example, I could not get the list sorted by Algorithm. It seems to be stuck trying to update itself.

There is no need to poll the remote server in order to sort the list of datasets being shown to the user. We sort the list according to what we currently know, we do not need to update the contents of the list, just sort them.

Please refactor this PR with this in mind

Samweli:
> There is no need to poll the remote server in order to sort the list of datasets being shown to the user.

Sure @ricardogsilva, at the moment we are not polling the remote server and sorting, I suspect the original sort implementation of [QSortFilterProxyModel](https://doc.qt.io/qt-5/qsortfilterproxymodel.html) is the cause. I'm going to customize it and see if it solve the issue.

luipir:
> There is no need to poll the remote server in order to sort the list of datasets being shown to the user

@ricardogsilva why do you think the code is doing api requests?

ricardogsilva:
> > There is no need to poll the remote server in order to sort the list of datasets being shown to the user
> 
> @ricardogsilva why do you think the code is doing api requests?

I didn't look at the code at all. In my mind, the only logical explanation for the slowness would be network IO. If that is not what the code is doing then it seems that there is likely a bug somewhere causing this delay. Lets fix it rather than adding progress messages to the user, as this type of sort should be a fast operation.

Samweli:
@luipir @ricardogsilva I have investigated the performance issue and found that the delay is might be caused by the [QSortFilterProxyModel.data](https://doc.qt.io/qt-5/qsortfilterproxymodel.html#data) method. see https://forum.qt.io/topic/106291/pyqt5-very-slow-qsortfilterproxymodel-not-customized
I implemented a different sort but still the performance didn't improve!
screenshot of time percent of QSortFilterProxyModel functions while sorting
running_time_percent_qtsortfilterproxymodel

luipir:
> @luipir @ricardogsilva I have investigated the performance issue and found that the delay is might be caused by the [QSortFilterProxyModel.data](https://doc.qt.io/qt-5/qsortfilterproxymodel.html#data) method. see https://forum.qt.io/topic/106291/pyqt5-very-slow-qsortfilterproxymodel-not-customized
> I implemented a different sort but still the performance didn't improve!

I'm not sure this post is related, just beacuse the guy is fetching remote db data and the object of the thread seems loosly related with internal staff of ProxyModel. BTW I thrust you that there is "strage" performance issue, strange in the sense tha the amount of data to sort seems not so high!
A possible bottleneck could be the data(...) method in original model but nothing is shoun in your above snapshot where I can neither see a proxymodel bottleneck :/

ricardogsilva:
so where do we stand on this @Samweli ?

Samweli:
> so where do we stand on this @Samweli ?

@ricardogsilva I'm still working on a fix, I ll update the PR when I have resolved the issue.

luipir:
@Samweli is it ok the performance issue? I see the code and it's not clear to me why you had to reimplement merge sort algorithm as solution for the performance issue. For me the code is ok, just I'm curious what was the element that reduced performaces.

Samweli:
> @Samweli is it ok the performance issue

@luipir the current changes doesn't have a performance issue, sorting is fast.


> For me the code is ok, just I'm curious what was the element that reduced performaces.

From my findings is that the `QSortFilterProxyModel sort` function was being called multiple times, i don't know why this happened yet, at the moment I have managed to stop the other extra calls.

I found another issue which I'm currently finding a fix for, when sorting if an item is highlighted by hovering the mouse, when the sorting finishes the item won't be updated until the item is un-highlighted.

luipir:
> when the sorting finishes the item won't be updated until the item is un-highlighted.

some event not emitted after sort?

Samweli:
> some event not emitted after sort?

@luipir maybe, but what event, the QSortFilterProxy `layoutChanged` signal is for sure emitted.

see the below screenshot, when sorting using `Name` finishes, all other items are updated except the one that was highlighted before and it is only updated after it is unhighlighted
highlight-issue-during-sorting

luipir:
> see the below screenshot, when sorting using `Name` finishes, all other items are updated except the one that was highlighted before and it is only updated after it is unhighlighted

Ha ok... I suppose it's not your problem I think. it's the delegate that probably leave the item in edit mode (the edit is activated with mouse over). Edit mode is the only way to allow accepting events.
Can you deselect any item before sorting?

Samweli:
> Can you deselect any item before sorting?

@luipir using [`QTreeView#reset`](https://doc.qt.io/qt-5/qtreeview.html#reset) make it work.

Samweli:
Rebased from the decision-trees branch

ricardogsilva (changes requested):
Sorting is fast now, which is great, thanks for digging into the issue!

However, there are still some problems.

1. Sorting by date does not seem to work correctly.
2. Sorting of other fields is not stable. If you press the sort button multiple times the list changes. This should not happen

Here is a small video demonstrating the problems mentioned above


https://user-images.githubusercontent.com/732010/118797356-69861e80-b894-11eb-92da-5484b428f684.mp4

Samweli:
> 1. Sorting by date does not seem to work correctly.

This might be because older datasets had their `creation_date` stored as a string and in recent datasets they are store as datetime objects(I might be incorrect). Have you tried removing older datasets and sort by date only recent ones.

> 2\. Sorting of other fields is not stable. If you press the sort button multiple times the list changes. This should not happen

@ricardogsilva @luipir what should be the expected behavior when sorting by Algorithm two datasets that have similar lets say 'Algorithm name' which one should be displayed first?

luipir:
> @ricardogsilva @luipir what should be the expected behavior when sorting by Algorithm two datasets that have similar lets say 'Algorithm name' which one should be displayed first?

second order by date?

ricardogsilva:
> > 1. Sorting by date does not seem to work correctly.
> 
> This might be because older datasets had their `creation_date` stored as a string and in recent datasets they are stored as datetime objects(I might be incorrect). Have you tried removing older datasets and sort by date only recent ones.

Is it not possible to make it work on both cases? If you use a string representation as the sort key it should work. Please try it out


> > 2. Sorting of other fields is not stable. If you press the sort button multiple times the list changes. This should not happen
> 
> @ricardogsilva @luipir what should be the expected behavior when sorting by Algorithm two datasets that have similar lets say 'Algorithm name' which one should be displayed first?

I agree with @luipir. Use the date as a second order sort parameter

Samweli:
> However, there are still some problems.

@ricardogsilva @luipir I have addressed the issues and updated the PR.

LDMP/models/datasets.py
```diff
@@ -548,6 +555,78 @@ def datasetById(self, id: str) -> Optional[Tuple[str, Job]]:
...
+
+    def __compare_dates(self, left, right):
+
+        if isinstance(left, datetime) and isinstance(right, str):
+            right = datetime.fromisoformat(right)
+        elif isinstance(left, str) and isinstance(right, datetime):
+            left = datetime.fromisoformat(left)
+        elif isinstance(left, str) and isinstance(right, str):
+            left = datetime.fromisoformat(left)
+            right = datetime.fromisoformat(right)
```
luipir, line 626: Am I wrong or can simplified with
```suggestion
        if isinstance(right, str):
            right = datetime.fromisoformat(right)
        if isinstance(left, str):
            left = datetime.fromisoformat(left)
```
in your case can multiply number of "if" in case of double string

luipir (changes requested):
other then this little modification, LGTM

LDMP/models/datasets.py
Samweli: you are right @luipir

luipir (approved):

luipir:
@ricardogsilva let me knwo if I can merge

luipir:
waiting for this merge to merge and rebase my branch about management of local processing... I'll to fix many conflicts

ricardogsilva (changes requested):
It seems to work nicely now!

However, in terms of functionality, I think it would be more useful for the user if the default sort would be the descending one. IMO the most useful field for sorting is the `date`. The others are nice to have, but they are like a bonus.  I think that the results should be sorted in such a way as that the most recent ones should show up at the top of the list.

Please make the default sorting work in such a way as the most recent items show on the top of the list and when the user checks the `Reverse` checkbox it is reversed (_i.e_ the most recent results show up on the bottom of the list).

ricardogsilva:
Lets merge this now and deal with reversing the default sort order later

Issue #398, opened 2021-05-03
Use compound run buttons only when needed


ricardogsilva:
The algorithms section shows a list of available algorithms.

Each algorithm shall only show a compoiund `Run` button when it is possible that it may be run both locally and remotely. Otherwise it shall show a simple button.

PR #427, opened 2021-05-16
Update algorithm run button

Samweli:
Fixes https://github.com/ConservationInternational/trends.earth/issues/398
Updates the algorithm run button to change according to the algorithm run mode state. 

Screenshot showing how run buttons will look accordingly to the respective algorithm run mode, the Land productivity algorithm
has been set to local and remote run modes for testing purposes.
update_alg_run_button

### DIFF
--- a/LDMP/gui/WidgetAlgorithmLeaf.ui
+++ b/LDMP/gui/WidgetAlgorithmLeaf.ui
@@ -79,7 +79,7 @@
   </property>
   <layout class="QVBoxLayout" name="verticalLayout">
    <item>
-    <layout class="QHBoxLayout" name="horizontalLayout" stretch="0,1">
+    <layout class="QHBoxLayout" name="horizontalLayout" stretch="0,0,1">
      <item>
       <widget class="QLabel" name="labelAlgorithmName">
        <property name="text">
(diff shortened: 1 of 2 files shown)


luipir:
LGTM but a general observation @Samweli Did you check if toolbutton can be configured to be rendered as pushButton? eg removing the right arrow? this would simplify the PR using a single UI with advantages in maintaining the code.

luipir:
e.g btn = QToolButton() btn.setStyleSheet('QToolButton::menu-indicator { image: none; }')

ricardogsilva (approved):
Looks good, thanks!

I'm +1 for merging, as soon as @luipir question is cleared and he also gives a green light

Samweli:
> LGTM but a general observation @Samweli Did you check if toolbutton can be configured to be rendered as pushButton?

Didn't thought about this @luipir, it seems it can be achievable.

Samweli:
> e.g btn = QToolButton() btn.setStyleSheet('QToolButton::menu-indicator { image: none; }')

@luipir that can do the trick. One issue though when using QToolButton you can't set a tooltip as it is generated automatically from the default action. I think it better to have two separate buttons though a QToolbutton for a both algorithm mode and a QPushbutton for a single algorithm mode.

luipir:
> @luipir that can do the trick. One issue though when using QToolButton you can't set a tooltip as it is generated automatically from the default action. I think it better to have two separate buttons though a QToolbutton for a both algorithm mode and a QPushbutton for a single algorithm mode.

then ok for me to merge

luipir (approved):

PR #430, opened 2021-05-21
Integration of local processings. Fixes #409

### LINKED ISSUE #409: Implement scanning of the local base dir in order to populate the datasets section

luipir:
Integrated (I hope) the availability to run all local executed processes.

**IMPORTANT NOTE: Because of refactoring. I strongly suggest to start with a clean base_data_directory folder to avois side effects in ase find json without expected values.**

The main architectural challenges were:
1) hook (and hide) output tab to setup a default set of output names
2) added some metadata (see metadata key in json) to store info useful to manage json as dataset: e.g. run_id, source and in some cases some other data inherited by the specific alg necessary to add dataset into the canvas
3) create a map of _source_ (name of GUI class in case of local processes) with the name of alg and and it's group. Necessary to store in base_data_directory and to shou in dataset gui.
4) add a _add_ to canvas method that manage band adds depending on type of product (there is no unique way). The best solution would be a registry of algs or datasets to delegate all type specific methods but need a more deep refactoring.

Limitations:
Respect remote processes, local processed datasets are added only at the end of process. The reason is to reduce the amount of modifications and leaved the workflow untouched. e.g. the local processing json is generated only at the end of process.
In this way data set is not present until the end of process, but it's process state is (as before) shown during it's process in the message bar.


LDMP/calculate.py
```diff
@@ -714,6 +756,76 @@ def check_overwrites(self):
         return True
 
 
+class CalculationHidedOutputWidget(QtWidgets.QWidget, Ui_WidgetCalculationOutput):
```
luipir, line 759: this class leave old CalculationOutputWidget unchanged for retro compatibility reasons and added a copy of the class tab where out names are set programmatically

LDMP/layers.py
```diff
@@ -328,19 +328,24 @@ def get_file_metadata(json_file):
...
 
 
-def json_serial(obj):
-    """JSON serializer for objects not serializable by default json code"""
-    if isinstance(obj, (datetime.datetime, datetime.date)):
-        return obj.isoformat()
-    raise TypeError("Type {} not serializable".format(type(obj)))
-
-
 def create_local_json_metadata(json_file, data_file, bands, metadata={}):
```
luipir, line 331: hooked method where Dataset is created (and related json)

LDMP/models/datasets.py
```diff
@@ -349,7 +381,68 @@ def add(self):
         Reused completly download method to try to use old code as back box and avoid introduce any
         changes. Datasets already downloaded will not downloaded again.
         """
-        self.download(datasets_refresh=False, add_to_map=True)
+        # do download if a Dataset generated from a remote Job
```
luipir, line 384: here all specific way to add a result in the canvas

ricardogsilva (changes requested):
I think you forgot to include some stuff that you probably have locally.

I cannot run the code, due to a missing `SortField` class, which does not seem to be defined anywhere.

```python
NameError: name 'SortField' is not defined 
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/qgis/utils.py", line 335, in _startPlugin
    plugins[packageName] = package.classFactory(iface)
  File "/home/ricardo/.local/share/QGIS/QGIS3/profiles/trends-earth-dev/python/plugins/LDMP/__init__.py", line 56, in classFactory
    from LDMP.plugin import LDMPPlugin
  File "/usr/lib/python3/dist-packages/qgis/utils.py", line 799, in _import
    mod = _builtin_import(name, globals, locals, fromlist, level)
  File "/home/ricardo/.local/share/QGIS/QGIS3/profiles/trends-earth-dev/python/plugins/LDMP/plugin.py", line 32, in 
    from LDMP.main_widget import get_trends_earth_dockwidget
  File "/usr/lib/python3/dist-packages/qgis/utils.py", line 799, in _import
    mod = _builtin_import(name, globals, locals, fromlist, level)
  File "/home/ricardo/.local/share/QGIS/QGIS3/profiles/trends-earth-dev/python/plugins/LDMP/main_widget.py", line 34, in 
    from LDMP.models.datasets import (
  File "/usr/lib/python3/dist-packages/qgis/utils.py", line 799, in _import
    mod = _builtin_import(name, globals, locals, fromlist, level)
  File "/home/ricardo/.local/share/QGIS/QGIS3/profiles/trends-earth-dev/python/plugins/LDMP/models/datasets.py", line 448, in 
    class Datasets(QObject):
  File "/home/ricardo/.local/share/QGIS/QGIS3/profiles/trends-earth-dev/python/plugins/LDMP/models/datasets.py", line 682, in Datasets
    field: SortField = SortField.DATE
NameError: name 'SortField' is not defined
```

It seems this `SortField` class would be defined in `models.datasets` but it is not there.

luipir:
Yes... I face the issue, and I'm working on it.

LDMP/calculate.py
```diff
@@ -65,6 +67,39 @@ class tr_calculate(object):
...
+local_scripts['DlgCalculateTCData'] = {
+    'group': 'total-carbon',
+    'display_name': 'Total Carbon (locally)',
+    'source': 'CalculateTCData'
+}
+local_scripts['DlgCalculateSOC'] = {
+    'group': 'SDG 15.3.1',
+    'display_name': 'Soil Organic Carbon (locally)',
+    'source': 'CalculateSOC'
+}
```
Samweli, line 102: Still I'll be best to have these defined in a separate file e.g. `data/scripts.json` and maybe add a `mode` property that will have `local` or `remote` values.

LDMP/calculate.py
```diff
@@ -714,6 +756,76 @@ def check_overwrites(self):
...
+        self.process_id = str(uuid.uuid4())
+        self.process_datetime = datetime.now()
+        self.process_datetime_str = self.process_datetime.isoformat()
+
+        # TODO: check where used "LDMP/output_dir" to avoid side effects
+
+        # path where to store result
+        base_data_directory = QSettings().value("trends_earth/advanced/base_data_directory", None, type=str)
+        if not base_data_directory:
+            return
```
Samweli, line 790: shouldn't we use a default base data directory instead of returning here?

LDMP/calculate.py
```diff
@@ -714,6 +756,76 @@ def check_overwrites(self):
...
+
+        processing_date_string = self.process_datetime.strftime('%Y_%m_%d')
+        group = get_script_group(self.subclass_name)
+        initial_path = os.path.join(base_data_directory, 'outputs', group, processing_date_string)
+        if not initial_path:
+            return
```
Samweli, line 796: same as above

Samweli (changes requested):

LDMP/calculate.py
luipir: I can do it, but I followed the same criteria as expressed times ago by @ricardogsilva regard the algorithm list configuration  file... no need to have an external code file because alg list is almost static => I didn't spent time to create conf file related with alg list.

luipir: > and maybe add a `mode` property that will have `local` or `remote` values.

adding a `mode` means change the actual workflow of processing e.g. change call to local or remote basing on this value e.g. more changes in calculate_xx.py basing on this value. I preferred to leave the workflow as before.

LDMP/calculate.py
luipir: `base_data_directory` is the default and have to be set otherwise nothing would work. I would set the default value in setting GUI

luipir:
@Samweli I answeered to your comments... if you agree with comments, please resolve them, or add more comments

LDMP/calculate.py
Samweli: I meant the `mode` property should be defined the in the file not on the class, same way the `group`, `display_name`  properties will be defined.

LDMP/calculate.py
luipir: > Yes... I face the issue, and I'm working on it.

hehehe... I wrongly removed the enum because was between two block of code that was marked as TODO: remove... I didnt 'expected some code was inserted in between :)

luipir:
I face this new error in sort funcion.
The reason is that can be some date with timezone set and other not. This is probably an error due the mix of old and datasets descriptors. fixing it just in case we face this kind of errors.
@Samweli take into account that datetimes are not alwasy comparable if a timezone is set only in one

Screenshot from 2021-05-24 12-10-12

LDMP/calculate.py
Samweli: > I would set the default value in setting GUI

@luipir does that mean the value will always be set?

Samweli:
> @Samweli take into account that datetimes are not alwasy comparable if a timezone is set only in one of them

right @luipir, what did you do to produce the error?

luipir:
> right @luipir, what did you do to produce the error?

when a dateime is set as string (e.g. is used now() for local processing) timezone is added in the string dump and obvioulsy maintained when converted to datetime.
The correct way to compare dates is always to assume they have timezone and do localize if not (or just localize and catch exception in case timezone is already set)

luipir:
@Samweli @ricardogsilva I assume pytz is a commonly installed python lib. Is it true in your installation?

ricardogsilva:
I don't have any windows machine at hand to try that out.

Would it be feasible to use only python's `datetime` for this?

We would store dates as utc internally and use them for this type of comparison, then show to the user in their local timezone. Something like:

```python
import datetime as dt
utc_date = dt.datetime.now(dt.timezone.utc)
local_date = utc_date.astimezone()
```

LDMP/calculate.py
luipir: > @luipir does that mean the value will always be set?

right! it is the folder where all datasets and jobs are stored

ricardogsilva:
I guess Qt's support for datetime may also work without having to depend on the user having `pytz` installed:

https://doc.qt.io/qt-5/qdatetime.html#details

LDMP/calculate.py
luipir: @Samweli sorry I can't understand, may you rephrase please?

luipir:
> Would it be feasible to use only python's `datetime` for this?

as far as I read... the only way to use only datetime is to assume all datetime are in the same tz (e.g. remove tz before comparison). This potentially can create problem in the future if someone want to import a dataset exported from someone else.

ricardogsilva:
> (e.g. remove tz before comparison)

I would rather make sure all datetimes we generate in the plugin use UTC instead of local time. If someone wants to import a dataset exported from somewhere else we would also assume UTC.

Alternatively, as I mention in the comment above, perhaps use Qt's datetime instead of Python's?

LDMP/calculate.py
Samweli: @luipir I think we need one place to define the all the algorithms properties not necessary at `data/script.json` we can have a `definitions.py` file instead where we define each alg with its corresponding properties including responsible class name.

for example, algorithm for land cover will have the following definition
 ```
"land-cover": {
        "script version": "1.0.3",
        "description": "Calculate land cover change indicator.",
        "group": "SDG 15.3.1",
        "mode": "local",
        "source": "CalculateLC",
        "display_name": "Land cover (locally)",
        "dialog": "DlgCalculateLC"
    }
```

**Review by ricardogsilva — CHANGES_REQUESTED** [review]
Base functionality of being able to run the local process and load the result to QGIS canvas seems to be working OK :smile_cat: 

However, we need some additional changes:

1. The result of running the local process is being stored on the filesystem with an obscure name (seems to be just an UUID). Lets provide a more descriptive name, in a similar way as the datasets that were generated from a remote process work - these are named with a mix of UUID and the name of the process that generated them.

2. The metadata present in the resulting `json` file needs to be more comprehensive. In order to be able to generate the dataset details dialog (see #407) we need to know:

   - [x] name of the algorithm
   - [ ] parameters used when running it - This must include also the input datasets that were used
   - [x] execution date
   - [ ] other files that were generated, which means the `json` and the `xls` file in this case

3. Ideally, the most important stats shown on the `xls` file would also be included in the `json` file, so that we would be able to show them on the dataset details dialog too. I'm speaking about the numbers shown on the sheet titled `SDG 15.3.1`. Can you also include these in the `json`? Otherwise we will need to read the `xls` file when generating the information for the dataset details dialog

File: LDMP/calculate.py (hunk 4 above)
Comment (luipir) [reply LDMP/calculate.py:102]: well the old code is bi-valent and it's an architectural problem at the time of refactoring.
Remote scripts are already classified in an external json: data/scripts.json BUT are present only remote scripts (probably associated to the scripts as present in GEE)
Local algorithms (no script concept is associated) are not classified at all and not all processes are local or remote processed... so local processing need a separate structure then remote `script`.
We can unify all, but at cost of a more heavy refactoring (at 6 day of the end of my consultancy). My guideline, with the time available to refactor, was use most of the code with lesser architectural impact... trying to use most of processing code as blackbox and modifying only interface.
I would say "wan't fix" in this phase other than having a clear instruction to do it.

**Comment by luipir** [comment]
> I would rather make sure all datetimes we generate in the plugin use UTC instead of local time. If someone wants to import a dataset exported from somewhere else we would also assume UTC.

I'll check where are the place where a datetime is set without time zone

**Comment by luipir** [comment]
> 1. The result of running the local process is being stored on the filesystem with an obscure name (seems to be just an UUID). Lets provide a more descriptive name, in a similar way as the datasets that were generated from a remote process work - these are named with a mix of UUID and the name of the process that generated them.

I'll check the impact to save in different filename format

**Comment by luipir** [comment]
> * parameters used when running it - This must include also the input datasets that were used

colud be difficult at dataset level where metadata are set at upper level in each algorithm... I've to check

**Comment by luipir** [comment]
> * execution date

is already there! are you talking about some other datetime value?

**Comment by luipir** [comment]
> 3\. Ideally, the most important stats shown on the `xls` file would also be included in the `json` file, so that we would be able to show them on the dataset details dialog too. I'm speaking about the numbers shown on the sheet titled `SDG 15.3.1`. Can you also include these in the `json`? Otherwise we will need to read the `xls` file when generating the information for the dataset details dialog

completly no idea... I never entered inside the algorithm logic. Please give me a prioritization.

**Comment by luipir** [comment]
> @Samweli @ricardogsilva I assume pytz is a commonly installed python lib. Is it true in your installation?
> sopmeone have a clean osgeo4w qgis installation to check if pytz lib is available by default?

checked in an olv VM I had and pytz is by default available

**Comment by luipir** [comment]
> checked in an olv VM I had and pytz is available by default

btw seems I can use datetime.datetime.now(datetime.timezone.ust) e.g. builtin modules

**Comment by luipir** [comment]
> 1. The result of running the local process is being stored on the filesystem with an obscure name (seems to be just an UUID). Lets provide a more descriptive name, in a similar way as the datasets that were generated from a remote process work - these are named with a mix of UUID and the name of the process that generated them.

this have a impact in the refactoring... actually the are two steps in any local process:
1) generating tiffs and/or ancillary data
2) generate local json descriptor

I hooked only the method generate json descriptor using the same naming convention used before (applied in dump method of dataset)

If I change this logic I should change all the naming logic in all local processes for all generated file by thy alg. This is heavy!
Another possibility is to rename all generated files after process (also modifying json content to be aligned).

Do we really need this right now @ricardogsilva ? the behaviour is different from GEE processes because different is the workflow!

NOTE. if I change only the generate json descriptor filename there is a heavy side effect... the produced images or xsls are not linked with filename and removed automatically because orphaned.

**Comment by luipir** [comment]
> f I change this logic I should change all the naming logic in all local processes for all generated file by thy alg. This is heavy!
> Another possibility is to rename all generated files after process (also modifying json content to be aligned).

checking another solution

**Comment by luipir** [comment]
> 1. The result of running the local process is being stored on the filesystem with an obscure name (seems to be just an UUID). Lets provide a more descriptive name, in a similar way as the datasets that were generated from a remote process work - these are named with a mix of UUID and the name of the process that generated them.

solved

**Comment by Samweli** [comment]
> The metadata present in the resulting `json` file needs to be more comprehensive. In order to be able to generate the dataset details dialog (see #407) we need to know:
> 
> * [x]  name of the algorithm
> * [ ]  parameters used when running it - This must include also the input datasets that were used
> * [x]  execution date
> * [ ]  other files that were generated, which means the `json` and the `xls` file in this case

right @ricardogsilva, @luipir how will we be able to access these information through the `dataset` object?

**Comment by luipir** [comment]
> right @ricardogsilva, @luipir how will we be able to access these information through the `dataset` object?

the unique way is to add into metadata... about this probably I need help (do you have time to work on this @Samweli )

- There is done exatly before any call of `create_local_json_metadata` . About "other files generated" is relative simple because their name is filled inside the hidden self.output_tab.output_summary (if output_tab is available) (see CalculationHidedOutputWidget.set_output_summary method)
- about **parameters used when running it** is really much more complicated... there is no homogeneous GUI, so each input data have be added differently for each algorithm dialog. BTW always right before the call of `create_local_json_metadata` 

I'm struggling with side effect (as announced) of "name of the algorithm" in file name (especially for remote tasks where response is not homogeneous).

**Comment by luipir** [comment]
> solved

@ricardogsilva facing side effect in remote GEE... working on that

**Comment by Samweli** [comment]
> do you have time to work on this @Samweli

@luipir I'm currently working on https://github.com/ConservationInternational/trends.earth/issues/407 and I did struggled to find a way to get the alg name, parameters and user inputs. One way or another I will have to work on it, as it need to populated the dataset details dialog with the information.

**Comment by luipir** [comment]
> find a way to get the alg name

yes there is a uniform way because change depending on response phase of remote GEE execution... this is an effect of separating Job from Dataset without possibility to align APIs.

You can infer where to get the alg name (displayed in Datasets list) from datasets_delegate

**Comment by luipir** [comment]
> @ricardogsilva facing side effect in remote GEE... working on that

seems solved now

**Comment by luipir** [comment]
rebased on latest decision-tree branch

**Comment by luipir** [comment]
> right @ricardogsilva, @luipir how will we be able to access these information through the `dataset` object?

@Samweli I can work on the remaining points. I would avoid to overlap if you already done something on this

**Comment by Samweli** [comment]
> > right @ricardogsilva, @luipir how will we be able to access these information through the `dataset` object?
> 
> @Samweli I can work on the remaining points. I would avoid to overlap if you already done something on this

@luipir yes that will be great, I haven't worked on anything related with this.

**Comment by luipir** [comment]
> 3\. Ideally, the most important stats shown on the `xls` file would also be included in the `json` file, so that we would be able to show them on the dataset details dialog too. I'm speaking about the numbers shown on the sheet titled `SDG 15.3.1`. Can you also include these in the `json`? Otherwise we will need to read the `xls` file when generating the information for the dataset details dialog

At code level is not simple at all other entering in the logic of the report (see make_summary_table methods) 
I strongly suggest to avoid to enter in any algorithm logic and add a button to open the xsls report (keep it simple :) 
Xsls file is listed in the metadata of the dataset descriptor => just use xdg_open to open with default xsls viewer

**Comment by luipir** [comment]
> @luipir yes that will be great, I haven't worked on anything related with this.

done

**Review by ricardogsilva — CHANGES_REQUESTED** [review]
Metadata about the result now includes all necessary fields, which is perfect for implementing #407, thanks.

I see a couple of minor issues still:

- [x] There is a pop up dialog being shown when the processing is done. Please get rid of it, as it seems unneeded. Information about where the outputs are stored can be consulted in the the details dialog, which shall be the result of #407 

- [ ] The final name of the results generated when I run the SDG15.3.1 algorithm is something like `{some-uuid}_LDNSummaryTable_{task-name}.{extension}`. I don't think `LDNSummaryTable` is a good name fragment to appear in there. I think you are probably using the name of the class? Please use the algorithm short description - I mean the one we are already showing on the algorithms list, which in this case shall be `Final SDG 15.3.1` instead

**Comment by ricardogsilva** [comment]
@luipir one additional thing I noticed while testing this.

I deleted my local Trends.Earth base dir and now I have some results being shown with no possibility to remove them from the T.E. datasets section:

Screenshot from 2021-05-27 15-45-44

These are no longer present on the local filesystem and I am assuming they are also not present on the remote GEE. However, I think the `delete` button should not be greyed out, because that means I cannot get rid of these entries now. Can you please make it so that the `delete` button is not disabled in this case?

**Comment by luipir** [comment]
> * The final name of the results generated when I run the SDG15.3.1 algorithm is something like `{some-uuid}_LDNSummaryTable_{task-name}.{extension}`. I don't think `LDNSummaryTable` is a good name fragment to appear in there. I think you are probably using the name of the class? Please use the algorithm short description - I mean the one we are already showing on the algorithms list, which in this case shall be `Final SDG 15.3.1` instead

ok... No ti's not the class name, but it's derived by class name and configured here:
https://github.com/luipir/trends.earth/blob/local_calculation/LDMP/calculate.py#L77

I wouldn't use spaces in filenames is it ok with "Final-SDG-15-3-1" and use "_" as field separator ?

**Comment by luipir** [comment]
> These are no longer present on the local filesystem and I am assuming they are also not present on the remote GEE. However, I think the `delete` button should not be greyed out, because that means I cannot get rid of these entries now. Can you please make it so that the `delete` button is not disabled in this case?

plase open a separate issue for that... It shouldn't be related with this PR

**Review by ricardogsilva — CHANGES_REQUESTED** [review]
Almost done

Perhaps I did not explain myself correctly with regard to the name of the results when running SDG 15.3.1 algorithm.

With these latest changes I see files named like this:

```
3772708f-f785-4886-bfcb-99db71c10225_final-sdg-15-3-1-summary-table_sdg15-3-1-3.json
```

Please remove `summary-table` from there, as it does not seem appropriate

luipir:
> These are no longer present on the local filesystem and I am assuming they are also not present on the remote GEE. However, I think the `delete` button should not be greyed out, because that means I cannot get rid of these entries now. Can you please make it so that the `delete` button is not disabled in this case?

sho0uld I leave it not greyed by default in any case... this should simplify logic and reduce use case errors

luipir:
> already suggested,

@ricardogsilva please review all other voices of the dictionary

ricardogsilva:
I don't want to change the other sources fro now. Is this ready for a final review from me?

Samweli:
> Also, @Samweli, I still don't see any reviews from your part, after nearly a week that this PR has been sitting in the queue. Please review ASAP. 
> 



@ricardogsilva I have already reviewed this PR and the issues have been addressed or explained. 

>
>This comment applies to any other PRs currently still in the queue. Please provide reviews in due time

I did provide the review in time, the issues I raised are already resolved. 

For the other PRs my reviews suggestions have been already added to the PRs. 


> Also, merge duty is on yourself, as we had previously agreed.
> 
I have been following up on the changes requested to be added on this PR, there were still somethings that @luipir was working on.

luipir:
@ricardogsilva @Samweli it's ready to review if @Samweli accepted my answers to his review questions (if so, please resolve conversation)

ricardogsilva:
@Samweli I am sorry. I see that you have indeed provided a review in a timely manner. It was my fault that I just skimmed over the comments and missed it somehow - this was obviously only my own fault, and not yours, so please accept this virtual apology :sweat_smile: 

Please feel free to merge this, if you are in agreement with the proposed changes.

Samweli:
> I see that you have indeed provided a review in a timely manner. It was my fault that I just skimmed over the comments and missed it somehow - this was obviously only my own fault, and not yours, so please accept this virtual apology

Alright @ricardogsilva

Samweli:
> it's ready to review if @Samweli accepted my answers to his review questions (if so, please resolve conversation)

right @luipir, I resolved the issues. One thing before merge, we have agreed not to convert the dates timezones when comparing them and instead we are now storing the datasets dates in UTC right?

luipir:
> right @luipir, I resolved the issues. One thing before merge, we have agreed not to convert the dates timezones when comparing them and instead we are now storing the datasets dates in UTC right?

I suppose we agreed the opposite or I misunderstand our conversation... time zone (converted to UTC) is necessary to compare dates. Do you have other solution @Samweli ?

luipir:
> we have agreed not to convert the dates timezones

these are the relevant comments:
by me:
_the only way to use only datetime is to assume all datetime are in the same tz (e.g. remove tz before comparison). This potentially can create problem in the future if someone want to import a dataset exported from someone else._
by @ricardogsilva 
_I would rather make sure all datetimes we generate in the plugin use UTC instead of local time. If someone wants to import a dataset exported from somewhere else we would also assume UTC._

I set all datetime adding or converting to UTC

Samweli:
> I suppose we agreed the opposite or I misunderstand our conversation... time zone (converted to UTC) is necessary to compare dates. Do you have other solution @Samweli ?

@luipir  No I don't have other solution, I didn't see the updates you made on the `compare_dates` function anymore, so I assumed that we don't alter the timezones during comparing dates.

Samweli:
> I set all datetime adding or converting to UTC

okay

Samweli (approved):

PR #431, opened 2021-05-21
Reduce dock widget startup instantiating GUI at plugin load. Fixes #390

### LINKED ISSUE #390: Opening the algorithm dock for the first time takes too long
When opening the algorithm dock there is a noticeable delay in the QGIS UI of about 1-2 seconds. This is probably being caused by some of the plugin's dialogues, which are being loaded into memory.

Please use a different initialization strategy in order to avoid locking up the main QGIS GUI

luipir:
Fixes #390

### DIFF
--- a/LDMP/main_widget.py
+++ b/LDMP/main_widget.py
@@ -68,6 +68,15 @@ def get_trends_earth_dockwidget(plugin):
     return _widget
 
 
+# instantiate calcluate callback container to reuse most of old code
+# do this before setting setupAlgorithmsTree
+# guis setup during plugin load to reduce GUI startup
+dlg_calculate_LD = DlgCalculateLD()
+dlg_calculate_TC = DlgCalculateTC()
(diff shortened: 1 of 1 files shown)


ricardogsilva (approved):
Much smoother now, thanks for the great work!

ricardogsilva:
@Samweli feel free to merge this after you review

luipir:
just a tech note... there were to way to implement this speedup

1) creating a callback method in charge to create runtime GUI instance. e.g. a new GUI instance every time
2) create one instance at startup and use it directly. This ti the solution adopted

1 and 2 differ from the fact that in 2 some setup can remain between run (for that value not set via settings) instead of setting up again every time.

for now this does not create any runtime or usability problem, other the need to setup default output filenames before run for local processing.

Samweli (approved):
Works nicely!

PR #440, opened 2021-06-01
Datasets details dialog

Samweli:
Fixes https://github.com/ConservationInternational/trends.earth/issues/407

Adds a dataset details dialog that shows datasets information along with the details of the algorithm used to create the dataset.

Screenshot showing the dataset details dialog
dataset_details_dialog


ricardogsilva:
@Samweli 

please fix the reported conflicts and let me know whether this is ready for a review

Samweli:
> @Samweli 
> 
> please fix the reported conflicts and let me know whether this is ready for a review

sure @ricardogsilva working on this.

Samweli:
@ricardogsilva I have fixed the conflicts.

ricardogsilva:
@Samweli this does not include the latest changes, i.e. it is missing `areaofinterest.py` module, so it does not run. Please update it

Samweli:
> @Samweli this does not include the latest changes, i.e. it is missing `areaofinterest.py` module, so it does not run. Please update it

right, done!

ricardogsilva (changes requested):
It is a good start, but I'd like to request the following modifications:

- [x] Make the dialogue be modal and provide it a parent. It should behave like the other dialogues

- [x] The dialogue is too tall and too narrow, make the default size wider, but smaller

- [x] Do not repeat information:

   - [x] task name shows up twice

   - [x] task status also shows up twice

- [x] Export as a zip should show its message in the details dialogue - otherwise it is difficult to see it, as the dialogue is likely in front of the QGIS main window

- [x] Add a button next to the task path to allow opening it with the file manager
- [x] Add the same icons used in the dataset list to the buttons used for deleting and loading a dataset.
- [x] Add an icon to the button that generates a zip
- [x] Zip file uses a name of `dataset.zip` - this is too generic. Name it similar to the name of the json file
- [x] Zip contents only include the tif output. Also put the json file in there
- [x] Zip file should go into a different directory - lets use a directory named `{trends_earth_base_dir}/exported`

Samweli:
@ricardogsilva thanks, I have updated the PR addressing all comments except for the below one

> * [ ]  Make the dialogue be modal and provide it a parent. It should behave like the other dialogues

I have had challenges in past and up till now, when I provide a parent to the dialog, the dialog can disappear without getting closed.

LDMP/jobs/mvc.py
```diff
@@ -26,6 +33,9 @@
 WidgetDatasetItemUi, _ = uic.loadUiType(
     str(Path(__file__).parents[1] / "gui/WidgetDatasetItem.ui"))
 
+WidgetDatasetItemDetailsUi, _ = uic.loadUiType(
```
ricardogsilva, line 36: Please put the details dialog code in its own file. The `jobs.mvc` module is good as it is.

LDMP/jobs/mvc.py
```diff
@@ -192,7 +202,37 @@ def updateEditorGeometry(
         editor.setGeometry(option.rect)
 
 
-class DatasetEditorWidget(QtWidgets.QWidget, WidgetDatasetItemUi):
+class DatasetActions:
```
ricardogsilva, line 205: Please do not overuse inheritance (and even more severe, multiple inheritance). IMO this is a fault of the original plugin. It makes the code harder to read and there is no clear advantage in this case.

In this case we do not need this `DatasetActions` class at all, as its methods can go directly in the `DatasetDetailsDialog` class.

Just to clarify, I am not against using inheritance, where it makes sense. There is currently no other code that needs the methods offered by `DatasetActions` so there is no need for this kind of abstraction.

LDMP/jobs/mvc.py
```diff
@@ -192,7 +202,37 @@ def updateEditorGeometry(
...
+        message_box.setIcon(QtWidgets.QMessageBox.Information)
+        result = QtWidgets.QMessageBox.Res = message_box.exec_()
+        if result == QtWidgets.QMessageBox.Yes:
+            log("About to delete the dataset")
+            for path in self.job.results.local_paths:
+                layers.delete_layer_by_filename(str(path))
+            manager.job_manager.delete_job(self.job)
+
+
+class DatasetEditorWidget(QtWidgets.QWidget, DatasetActions,  WidgetDatasetItemUi):
```
ricardogsilva, line 235: Please rename this to `DatasetEditorDialogue` and make it a subclass of `QtWidgets.QDialog`.

Also, as mentioned above:

- put this code in another file, not on `jobs.mvc`
- `DatasetActions` should not be a superclass to this one. Merge it into this

LDMP/jobs/mvc.py
```diff
@@ -265,29 +305,91 @@ def __init__(self, job: models.Job, parent=None):
...
+
+        self.load_job_details()
+
+        self.bar = qgis.gui.QgsMessageBar()
+        self.bar.setSizePolicy(
+            QtWidgets.QSizePolicy.Minimum, QtWidgets.QSizePolicy.Fixed
+        )
+        self.layout().addWidget(self.bar, 0, 0, alignment=QtCore.Qt.AlignTop)
+
+    def __delete_dataset(self):
```
ricardogsilva, line 356: Please do not use this kind of naming scheme for methods. Don't use double underscores for naming methods that are not Python special methods. I am aware some devs do it to signal that this is a private method. The usefulness of this in a language like Python can be debated, but for brevity, for this project, lets just no use it.

LDMP/jobs/mvc.py
```diff
@@ -265,29 +305,91 @@ def __init__(self, job: models.Job, parent=None):
...
+        self.delete_dataset()
+        self.close()
```
ricardogsilva, line 358: make this class be a `QDialog` and use `self.accept` instead of `self.close`

LDMP/jobs/mvc.py
```diff
@@ -265,29 +305,91 @@ def __init__(self, job: models.Job, parent=None):
...
+
+        manager.job_manager.exports_dir.mkdir(exist_ok=True)
+
+        try:
+            zipped_file = f"{str(manager.job_manager.exports_dir)}/" \
+                          f"{self.job.params.task_name}.zip"
+            with ZipFile(zipped_file, 'w') as zip:
+                for file in files:
+                    zip.write(file, os.path.basename(file))
+        except Exception:
```
ricardogsilva, line 379: Always avoid using the generic `Exception` in a try block. Check which exceptions can happen and write code to guard against those.

If we ever get some unforeseen exception we fix the code

LDMP/jobs/mvc.py
```diff
@@ -265,29 +305,91 @@ def __init__(self, job: models.Job, parent=None):
 
     def show_details(self):
         log(f"Details button clicked for job {self.job.params.task_name!r}")
+        result = DatasetDetailsWidget(self.job).exec_()
```
ricardogsilva, line 308: make the qgis main window become the parent of this dialogue.

ricardogsilva (changes requested):
Functionality-wise it seems to be working OK.

Please adjust the code according to my comments and then I'll be +1 for merging

ricardogsilva (approved):
LGTM!

PR #442, opened 2021-06-01
Add settings for download and remote polling frequency

### LINKED ISSUE #374: Allow modifying the refresh frequency of the remote GEE server
Introduce new settings (under the `advanced` section) to allow:

1. Enabling/disabling automatic polling of the remote
1. Configuring the frequency of the refresh operation. It shall make use of a widget that accepts only non-negative integer values, which are to be interpreted as seconds. Make it default to `60`, meaning execute a new refresh every minute.

Here is a mockup of how the settings shall look like:

geo-ldn-settings-dialogue-advanced-section

### NOTE

Refreshing the list of datasets actually means two different things:

1. Polling the remote server for updates on datasets being generated and possibly initiate their download
1. Scanning the `base dir` for the presence of new datasets. - this shall not be controllable by the end user. The work described in this issue does not apply here.

### LINKED ISSUE #376: Implement automatic download of remote datasets

ricardogsilva:
This PR features an extensive refactor of the plugin internals in order to implement correct behavior for polling both local filesystem and the remote GEE server.

It also implements changing both the remote server polling frequency and also the automatic download of finished results.

Although the work for enabling local processes is not complete yet, I think it is best to merge this in its current state, in order to not block further work.

Fixes #376 
Fixes #374


ricardogsilva:
@Samweli 

This is not done yet, but you can start studying the changes. Brief overview:

- There is now a `conf` module, which has a `settings_manager`. This is the way to get settings now. Also, all of our settings keys should be represented by the `Setting` enum (unless I forgot to add some). This module also has the `ALGORITHM_TREE` object, which is where the configuration of the Trends.Earth algorithms is stored.

- Added the `jobs.manager.job_manager`, which is a manager of jobs and results. It is able to update local state by checking the filesystem. It is also able to refresh the local state from the remote server. It is also able to download job results. This replaces the previous `jobs.py` and `datasets.py` stuff. We are not storing anything about the current status of jobs and results on the qgis settings anymore, everything is cached on the filesystem and updated into the in-memory internals of the `job_manager` periodically, by calling its `refresh_local_state()` or `refresh_from_remote_state()` methods.

- Added a more type hint friendly way to parse job responses from the server and extended it to be able to serialize/deserialize it to local disk with ease. Check the `jobs.models.Job` class.

- Refactored a bunch stuff related to the GUI, like adding type hints for widgets, renaming them, etc.

There is still some work to do, but I am confident in having it ready soon.

ricardogsilva:
@Samweli 

I'm going to go ahead and merge this, even if local processes are not working OK yet. I  want you to start adapting your pending PRs to work with this ASAP.

We can add support for local processes in a subsequent change

Samweli:
> @Samweli
> 
> I'm going to go ahead and merge this, even if local processes are not working OK yet. I want you to start adapting your pending PRs to work with this ASAP.
> 
> We can add support for local processes in a subsequent change

right @ricardogsilva

Issue #438, opened 2021-06-01
Split sub-indicator for land cover into two


ricardogsilva:
SDG 15.3.1 Land Cover sub-indicator has two different execution modes

- Run remotely with default data
- Run locally with user-supplied data

As such we need to make sure the `run` button in the algorithm list is compound and presents two options.

- `run with custom data` - this shall be the default action. It shall open an algorithm dialog that allows the user to import data for the initial and final years and then define the degradation matrix.
- `run with default data` - this shall open an algorithm dialog that allows using the ESA CCI-LC dataset.

This means we will need two different dialogs, one only shows the options for importing local data, the other only shows the options for using ESA CCI-LC dataset.

Issue #439, opened 2021-06-01
Split sub-indicator for Soil Organic Carbon into two


ricardogsilva:
Similarly to the Land Cover sub-indicator (see #438) we also need to split the SOC sub-indicator into two different execution dialogs.

One of them shall use locally imported datasets while the other shall use the remote data.

PR #453, opened 2021-06-17
Land cover and soil carbon algorithm dialogs update

Samweli:
Fixes https://github.com/ConservationInternational/trends.earth/issues/438 and https://github.com/ConservationInternational/trends.earth/issues/439

Updates the land cover and soil carbon algorithms to show different dialog setup options accordingly to the current algorithm run mode.

Screenshot showing the land cover and soil carbon dialog new look
land_cover_soil_new_look


ricardogsilva (approved):
I did some extra commit in order to fix some stuff that was not working:

- region name did not update itself when the user selected a different region from the algo execution dialog
- algorithm that calculates all three sdg15.3.1 indicators at once was not running

I also split the land cover setup widget into two widgets and adapated the code

Issue #455, opened 2021-06-18
close dataset details dialogue after adding dataset to QGIS canvas


ricardogsilva:
When a user is looking at the dataset details dialogue and then presses the `Load dataset` button, the dialogue should close in order to allow the user to focus on the QGIS canvas

PR #460, opened 2021-06-18
Close the datasets details dialog when dataset is being loaded into QGIS

Samweli:
Fixes https://github.com/ConservationInternational/trends.earth/issues/455


ricardogsilva (approved):
short and sweet, LGTM!

Issue #456, opened 2021-06-18
Button to add dataset to QGIS canvas should be disabled for datasets which are not loadable


ricardogsilva:
Some of the existing datasets cannot be loaded onto QGIS canvas (e.g. they are .xlsx files) - this is true for example for the result of the `total-carbon-summary` algorithm.

The dataset list already has the `add to QGIS canvas` button disabled for these. Implement it also for the dataset details dialogue

PR #461, opened 2021-06-18
Disable add dataset button when the dataset can't be loaded

Samweli:
Fixes https://github.com/ConservationInternational/trends.earth/issues/456

The datasets are unloadable when they have not been downloaded or don't have local paths.


ricardogsilva (approved):
looking good, thanks!

PR #501, opened 2021-08-21
Complete uic transition

elpaso:
Remove all pyuic and resource dependencies to be able to run the plugin from source tree without compiling UI and resource files.

Also imports PyQt from qgis as recommended by QGIS development guidelines.


elpaso:
CC @azvoleff

Issue #518, opened 2022-01-11
Metadata editing workflow


timlinux:
![image](https://user-images.githubusercontent.com/178003/148951183-98b4d77b-246f-4b69-9f40-38191350803d.png)

![image](https://user-images.githubusercontent.com/178003/148951277-39d233e1-29c2-4fe6-ba4e-f27b1eb52ed1.png)

![image](https://user-images.githubusercontent.com/178003/148951315-4e9a5d07-a960-4bb9-9a17-29bdef81c9cb.png)

Add metadata button on the form above

Once the layer is already loaded in the project  we can open the Dataset details window and then have a new button that will take us to the full QGIS layer properties dialog with the metadata tab opened

timlinux:
If we can not match the state to the iso metadata model, dont try to store in the metadata

timlinux:
![image](https://user-images.githubusercontent.com/178003/148955047-b23911f4-04de-414d-bd14-10268774c73c.png)


We already have an export as zip button, but Alex B needs to modify the function to export the metadata into an ISO doc and embed it in the zip.

timlinux:
![image](https://user-images.githubusercontent.com/178003/148955244-3ad96f21-97bf-40b8-b299-101bf63ab810.png)


This existing document also needs to be refactored so that the data storage for the metadata is stored in the QGIS layer properties.

timlinux:
So to summarise:

Phase 1:

 Alex B to hook in the layer properties so that layer metadata is stored in the QMD / QGIS layer metadata:
 Then add functionality so that we can transform the QGIS metadata into an ISO doc (look in metacat bridge plugin for inspiratio)
 Lastly when the layer properties export to Zip buttion is pressed, the ISO doc should be bundled in
 
 
 Phases 2: 
 
 Alex Z will be including extra metadata in e.g. json doc with the layer payload that comes from google EE
 Alex B to parse that and include as part of layer metadata and surface in the above T.E layer properties dialog

alexbruy:
Splitting Phase1 into smaller task for better manageability:

- [ ] Update all dialogs for loading custom datasets (Soil Organic Carbon, Land Cover, Land Productivity) by adding a "Metadata" button to them. Clicking on this button will open a metadata editor identical to QGIS metadata editor where user can enter metadata of the dataset. Metadata is stored in the QGIS layer properties and in XML file.
[...]

PR #527, opened 2022-01-25
Dataset metadata support

alexbruy:
This is an initial implementation of the dataset metadata editing workflow.

1. On a custom dataset import user can edit metadata and they will be stored together with the imported dataset. The same metadata will be propated to imported raster layer.
2. A new "Metadata" button were added to the "Dataset details" dialog and to the individual dataset items in the "Datasets" tab. Clicking on this button open metadata editor for a dataset.
3. When job results are downloaded plugin automatically creates metadata for dataset and individual layers. 
4. When individual layers are added to QGIS it is possible to edit their metadata using QGIS metadata editor from layer properties.
5. Exporting dataset will also export dataset and layers metadata in ISO XML format. For this functionality a new dependency were to LXML library were added (this library available on all major platforms, so I assume it is not an issue).


alexbruy:
As discussed we use simple metadata editor.
metadata

boc84:
> 1. On a custom dataset import user can edit metadata and they will be stored together with the imported dataset. The same metadata will be propated to imported raster layer.

Would the metadata editor apply to both imported raster and vector layers or just raster?

alexbruy:
> Would the metadata editor apply to both imported raster and vector layers or just raster?

Metsadata will be stored only for imported raster. Original files are not affected.

LDMP/metadata.py
```diff
@@ -0,0 +1,212 @@
...
+    out_dom = transform(in_dom)
+
+    s = '<?xml version="1.0" encoding="UTF-8"?>\n' + ET.tostring(out_dom, pretty_print=True).decode()
+    with open(temp_file, 'w', encoding='utf8') as f:
+        f.write(s)
+
+    return Path(temp_file)
+
+
+def init_datatset_metadata(job: Job, metadata: qgis.core.QgsLayerMetadata = None):
```
azvoleff, line 52: datatset -> dataset

LDMP/metadata_dialog.py
```diff
@@ -0,0 +1,220 @@
...
+        self.btn_add_default_category.setIcon(
+            QtGui.QIcon(os.path.join(ICON_PATH, 'mActionArrowRight.svg'))
+        )
+        self.btn_remove_category.setIcon(
+            QtGui.QIcon(os.path.join(ICON_PATH, 'symbologyRemove.svg'))
+        )
+
+        self.layer = None
+        self.metadata = None
+        self.default_categories = [tr("Farming"), tr("Biota"),
```
azvoleff, line 48: Assume these categories come from the ISO standard?

azvoleff:
Can you make relative to https://github.com/ConservationInternational/trends.earth/commit/34766ef8196ee1c7f8d02edc4a9af888e5f7f610? We should use that as the base for any further pull requests / branches

LDMP/metadata_dialog.py
alexbruy: They are coming from the document attached here https://github.com/ConservationInternational/trends.earth/issues/518#issuecomment-1017538060. And these are also standard categories used in the ISO standard.

alexbruy:
@azvoleff rebased.

azvoleff:
In testing this locally I got a python error after closing the dataset window when editing metadata for an existing dataset:

```
2022-02-17T08:13:01     WARNING    Traceback (most recent call last):
              File "D:/Documents and Settings/azvoleff/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\LDMP\jobs\mvc.py", line 388, in 
              action.triggered.connect(lambda _, x=file_path: self.show_metadata(x))
              File "D:/Documents and Settings/azvoleff/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\LDMP\jobs\mvc.py", line 347, in show_metadata
              dlg = metadata_dialog.DlgDatasetMetadata(self)
              File "D:/Documents and Settings/azvoleff/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\LDMP\metadata_dialog.py", line 27, in __init__
              super(DlgDatasetMetadata, self).__init__(parent)
             RuntimeError: wrapped C/C++ object of type DatasetEditorWidget has been deleted
```

Also the ability to edit metadata for the various types of data within each dataset isn't working quite how I'd expected. Right now it allows editing metadata for each component within the overall dataset vrt - so for example for a SDG 15.3.1 analysis you get:

image

What I'd expected was for this to be tied to the dataset type field, so that you would instead be able to edit metadata for the various subcomponents that show up when, for example, you select the "Select specific layers from dataset to add to map" screen:

image

However, in looking at the above it is a VERY long list (and some datasets have even _more_ band within them). The above list is produced using the `name` field accompanying each `band` in the `bands` dictionary that accompanies each dataset. I suggest we (my team) add another key to that dictionary, to assign a `category` to each band. Metadata should then be assigned by category (which would result in only probably 3-4 (maximum) metadata entries per dataset. For backwards compatibility, if that category key is missing, then there should be only 1 metadata entry for the entire dataset.

Would the above proposal work? I don't think it would mean changing much in the code, the code would just need to look for and use that category key for grouping rather than grouping by datatype as is now done.

alexbruy:
> Also the ability to edit metadata for the various types of data within each dataset isn't working quite how I'd expected. Right now it allows editing metadata for each component within the overall dataset vrt 

That's true, it works for a whole raster, not for individual bands inside raster. Otherwise it won't be possible to edit metadata from QGIS layer properties, as QGIS (and other tools) works with a raster as a whole.

azvoleff:
Merged into develop by c1842bb4f

PR #536, opened 2022-01-27
Add new images and training modules for UNCCD reporting and web-versi…

azvoleff:
@MLNoon - when I run "invoke docs-build" I get a large number of errors regarding images that can't be found. Do you see these errors on your side?


azvoleff:
Fixed in ec3dfbb3c1edabffcbb9ab643a38636e321699f7

PR #553, opened 2022-02-15
Special areas digitizing (false positive/false negative)

alexbruy:
Add support for digitizing special areas, namely false positives/false negatives.

To create new special area dataset one should use new "Create…" button at the bottom of the Trends.Earth dock, for now only false positives/false negatives are supported.
зображення
Clicking on the menu item will add a new dataset to the list of available datasets. For better usability there is a filter button at the top-right corner of the Trends.Earth dock where one can choose between showing all datasets, only raster datasets, only vector datasets (special areas)
зображення
Two buttons in the special area item allow to add vector layer to canvas for viewing and editing respectively.

Once added to canvas, layer uses predefined styling
зображення

There are two new buttons added to the plugin toolbar for digitizing special areas.
зображення
First button allows manual digitizing and displays area of the digitized feature in the top-right corner of the canvas
зображення
Second button provides a buffer digitizing tool, after specifying center of the area one can create a circular buffer around center point. Buffer can be created manually or by entering buffer radius in km in the widget in the top-right corner of the canvas or entering a desired area of the buffer in sq.km (the latter requires switching to area mode)
зображення
Also any added to canvas layer can be modified with QGIS native digitizing tools.

When feature digitized user presented with custom attribute form where they can fill in all necessary attributes of the special area.
зображення
Some of the attributes pre-polulated, e.g. feature area, UUID, source (either "manual digitizing" or "buffer tool"). Other attribute values should be entered manually or selected from the list of predefined values.

Requires ConservationInternational/trends.earth-schemas#2.


azvoleff:
See #567

PR #561, opened 2022-02-24
Dataset Report Generation

gkahiu:
A few notes on this PR:

1. Kindly check and confirm that I have included the key algorithms in the template configuration file in `LDMP/data/reports/templates.json`
2. I was unable to test the generation of multi-dataset reports since the LDN summary algorithm kept on terminating unexpectedly and removing the job from the Datasets window. I will file a separate issue on this. Once this has been resolved then I will update the PR accordingly.

Looking forward to your feedback!

Cheers,

John


gkahiu:
Updated PR includes support for land cover charts and auto masking of basemap.

azvoleff:
Thanks John, this is great, and exciting to see. Merging this into develop so we can test more easily with the other new metadata code, etc., and take it from there.

PR #589, opened 2022-03-28
fix broken dataset removal (fix #578)

### LINKED ISSUE #578: Data set deletion tool inconsistent
With the additional dataset types, we occationally get a python error or a NOP when the user attempts to delte a dataset. Please investigate and fix.

alexbruy:


azvoleff:
This seems to partially work - if I delete a dataset it removes the files associated with that dataset, and adds a json file for that job to the `deleted-jobs` folder. However, then on restarting QGIS the deleted job shows back up in the Datasets window (I assume when the plugin again polls our server to see what existing jobs that are for a particular user ID). Can you check that the code is properly checking the `deleted-jobs` folder and excluding those jobs from the datasets tab?

alexbruy:
> Can you check that the code is properly checking the deleted-jobs folder and excluding those jobs from the datasets tab?

Seems this also has never worked properly, as the list of the deleted jobs was not populated until the first download.

PR #590, opened 2022-03-28
add menu to load rasters for special areas (fix #586)

### LINKED ISSUE #586: Need to be able to re-load the layers user used for the false/positive negative tool
After experimenting with the false positive/negative tool, I realize we need a way for users to easily add back to their map the raster layers that they referenced while generating the layer. One way to do this would be to modify the button used to add vector layers to the map. Right now it is a simple pushbutton:

image

Can we make it a button with a dropdown, similar to what we do with raster layers:

image

In this case the user could either "Add vector to map", or "Add reference rasters to map" (to load the rasters they selected while digitizing the layer). We'd also need to raise a message to alert users in the case of those rasters no longer being available.

alexbruy:


azvoleff:
I initially get the button with an option to add the raster layers to the map, but then after I edit the layer the buttons disappear, and I instead have a greyed out download button:

image

alexbruy:
Looking at the code, this happens because for some reason plugin incorrectly parses job file and threats vector datasets as file datasets.

timlinux:
@azvoleff notes that the issue is probably related to 17c8b797f4014f76214ca462233626e13947113f

azvoleff:
I fixed the issue with special areas datasets related to serializing/deserializing vector JSONs in https://github.com/ConservationInternational/trends.earth-schemas/commit/d5b91c21933a569e335a4f869833ca9648ef3601 (so you'll need to update trends.earth-schemas). But this issue still remains - I get no charts from the tool.

PR #591, opened 2022-03-28
drop existing style from the GPKG (fix #587)

### LINKED ISSUE #587: Popup regarding style "already exists in the database" for false pos/neg tool
When adding a new false pos/negative layer, I get the following message when I start editing the layer:

image

Why does this message pop up? Can we avoid having the user see this message? It pops up up 8 times for me if I click "yes" each time.

Then when I try to add a new polygon I get:

`2022-03-28T08:59:58     WARNING    Traceback (most recent call last):
              File "D:\Documents and Settings/azvoleff/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\LDMP\maptools.py", line 193, in cadCanvasMoveEvent
              self.widget.set_area(g.area() * f)
             AttributeError: 'NoneType' object has no attribute 'set_area'`

alexbruy:
Old QGIS versions show popup when style saved under the same name, so we have to drop it before saving.

### DIFF
--- a/LDMP/layers.py
+++ b/LDMP/layers.py
@@ -987,4 +987,8 @@ def set_default_value(v_path: str, field: str, r_path: str, band: int, v: int):
 
     idx = layer.fields().lookupField(field)
     layer.setDefaultValueDefinition(idx, QgsDefaultValue("calculate_charts('{}', {}, {})".format(r_path, band, v), True))
+    exists, msg = QgsProviderRegistry.instance().styleExists('ogr', layer.source(), 'false_positive')
+    if exists:
+        QgsProviderRegistry.instance().deleteStyleById('ogr', l.source(), '1', msg)
+
     layer.saveStyleToDatabase('false_positive', '', True, '')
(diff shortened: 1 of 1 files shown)


azvoleff:
Looks like `styleExists` was added in 3.24. On 3.22 I get:

```
2022-03-28T15:15:45     WARNING    Traceback (most recent call last):
              File "D:\Documents and Settings/azvoleff/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\LDMP\jobs\mvc.py", line 462, in edit_layer
              layers.set_default_value(
              File "D:\Documents and Settings/azvoleff/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\LDMP\layers.py", line 990, in set_default_value
              exists, msg = QgsProviderRegistry.instance().styleExists('ogr', layer.source(), 'false_positive')
             AttributeError: 'QgsProviderRegistry' object has no attribute 'styleExists'
```

Can we make this check compatible with 3.22?

alexbruy:
Should work with old QGIS versions now.

azvoleff:
On 3.22, with this fix the error goes away, but now I see the popup 8 times again

alexbruy:
Ok, used another approach. Hope it will work for you.

azvoleff:
Thanks - that works!

Issue #583, opened 2022-03-25
Python error on buffering an areaofinterest


azvoleff:
Under the settings tool apply a buffer to any area of interest:

![Inkedqgis-ltr-bin_jG99rEPj4N_LI](https://user-images.githubusercontent.com/107753/160129787-d62fb054-8075-41b2-9163-8691e0e4faea.jpg)

Then open any algorithm and try to run it using that area of interest, and a Python error will come up:

```
2022-03-25T09:26:11     WARNING    Traceback (most recent call last):
              File "D:\Documents and Settings/azvoleff/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\LDMP\calculate_urban.py", line 241, in showEvent
              self.combo_layer_urban_series.populate()
              File "D:\Documents and Settings/azvoleff/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\LDMP\data_io.py", line 1446, in populate
              aoi = areaofinterest.prepare_area_of_interest()
              File "D:\Documents and Settings/azvoleff/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\LDMP\areaofinterest.py", line 540, in prepare_area_of_interest
              ret = area_of_interest.buffer(float(buffer_size))
              File "D:\Documents and Settings/azvoleff/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\LDMP\areaofinterest.py", line 376, in buffer
              to_aeqd, qgis.core.Qgis.TransformDirection.ReverseTransform
              File "C:\OSGeo4W\apps\Python39\lib\enum.py", line 405, in __getattr__
              raise AttributeError(name) from None
             AttributeError: ReverseTransform
             
```

I am running Qgis 3.22 - based on a quick search looks to be related to a [change in Qgis 3.22](https://qgis.org/pyqgis/3.22/core/QgsCoordinateTransform.html#qgis.core.QgsCoordinateTransform.TransformDirection)?

azvoleff:
@timlinux - who would be best top assign this one to?

azvoleff:
Fixed by #605

PR #605, opened 2022-04-05
Fix Geometry Transformation Enum in AOI Buffer Operation

gkahiu:
Fix for #583


azvoleff (approved):

azvoleff:
Thanks - looks great.

Issue #585, opened 2022-03-25
Charting of false pos/negative datasets isn't working


azvoleff:
I tried a test for Botswana (admin 2 region "North East"). When I enter a polygon I assume I should get a chart. I just get a blank chart though:

![image](https://user-images.githubusercontent.com/107753/160156950-cb8b5ae2-c0f7-4004-9a2f-1a141eb3e538.png)

Possible I am doing something wrong? This is on [25ae4c8](https://github.com/ConservationInternational/trends.earth/commit/25ae4c8b48258c6bf814f31425a18b7dd2cfada7)

azvoleff:
Also I get a Python error on trying to save the vector:

![image](https://user-images.githubusercontent.com/107753/160157834-04355efb-8786-40e7-9065-8fe915f6bc50.png)

alexbruy:
Look like your rasters does not have valid pixel values: -1, 0 and 1 for degraded, stable and improved respectively.

azvoleff:
That doesn't seem to be the issue. There actually were a few layers that used different coding (forgot one is a percentage, and the other 1-5). But I fixed that issue in the [vector_charting branch](https://github.com/ConservationInternational/trends.earth/tree/vector_charting), and I still don't get any charts - just blank charts without any error message. It seems the `calculate_charts` function never runs.

alexbruy:
`calculate_charts` runs fine and produce correct results, but these values are not saved to the attribute table for some reason. Without the values in the fields it is not possible to display charts

azvoleff:
I fixed the issue with special areas datasets related to serializing/deserializing vector JSONs (that came up in https://github.com/ConservationInternational/trends.earth/pull/590 in https://github.com/ConservationInternational/trends.earth-schemas/commit/d5b91c21933a569e335a4f869833ca9648ef3601 (so you'll need to update trends.earth-schemas). But this issue still remains - I get no charts from the tool.

azvoleff:
The charting still isn't working on the latest version (even after merge of #600 and fix noted above for the issue noted in #590). Right now no charts come up during editing, and on clicking the edit button to stop editing a layer, I get:

![image](https://user-images.githubusercontent.com/107753/162205271-157d843c-63cf-4a8f-be63-3ab47443ea0b.png)

PR #607, opened 2022-04-05
Rework logic behind add/edit buttons for special areas (fix #600)

### LINKED ISSUE #600: Fix edit vs add for vector layers
Right now [`edit_special_area_layer`](https://github.com/ConservationInternational/trends.earth/blob/develop/LDMP/jobs/manager.py#L545) and [`display_special_area_layer`](https://github.com/ConservationInternational/trends.earth/blob/develop/LDMP/jobs/manager.py#L541) do the same thing.

- `edit_special_area_layer` should only enable/disable editing of a special area layer that is already present on the map. So it should be greyed out if that particular dataset is not on the map, and then work to enable/disable editing if the layer **is** present on the map
- `display_special_area_layer` works as expected and doesn't need to be changed (i.e. it supports adding a special area layer to the map, and/or adding the supporting rasters for that layer to the map)

alexbruy:
Initially approved and implemented behaviour has to be changed as follows:
- add button should add layer
- edit button should toggle editing of the previously added layer
- edit button state should depend on whether layer present in the layer tree.

Fixes #600.


azvoleff:
Thanks @alexbruy. This change is enabling/disabling the wrong button. With this change if I add a layer to then the add layer button then gets disabled:

image

That button should remain enabled (so that a user can also add the associated raster layers to the map, for example). As I noted on #600, the needed change are to ensure that:

>`edit_special_area_layer` should only enable/disable editing of a special area layer that is already present on the map. So edit_tb for a vector dataset should be greyed out if that particular dataset is not on the map, or, if the layer is present on the map, then edit_tb should support enabling/disabling editing of that layer.

The button to add the vector layer to the map should always be enabled, regardless of whether the layer is present on the map or not.

alexbruy:
Sorry, sleepless night. Fixed.

azvoleff:
Hmm. Still not quite right. I can create a layer and add it to the map with the add button, but the edit button is still disabled. If I restart qgis then the edit button is enabled, but still not working quite right, as it is enabled regardless of whether the layer is on the map or not.

azvoleff:
That works, but the charting still doesn't (as noted on #585). Closing this to move the discussion on that issue there.

Issue #580, opened 2022-03-22
Manage of classes for land cover change targets


timlinux:
![imagem](https://user-images.githubusercontent.com/178003/159504688-4c3bf003-e5c2-459f-9bda-45eebcc2bb5c.png)

User needs to be able to change the number and names of cover classes similar to this tool

![imagem](https://user-images.githubusercontent.com/178003/159504885-6ff56a1f-7af2-4707-9847-eaa1d3ba6218.png)

After changing the user should be able to save and load their cover definitions

timlinux:
If the user adds more classes, they should be nested under the original 7 classes. The original 7 classes will be fixed and user may only add nested classes. Column names for this tool would differ from the aforementioned tool

![imagem](https://user-images.githubusercontent.com/178003/159505698-93e0ecbd-c0a7-4ea9-bb22-6d9b3d04244f.png)

![imagem](https://user-images.githubusercontent.com/178003/159506153-41b0c559-cb0a-4da7-9a8c-f6eabb580672.png)

timlinux:
@alexbruy also note that the user should be able to nominate a colour for each of the new classes, so that e.g. the charting from John can pick up the color defined here.

timlinux:
@alexbruy if you have not already started on this item, please let us know and we can assign it to @gkahiu

gkahiu:
@azvoleff, just to make sure that I understand correctly:

1. We should allow adding of nested classes (in the window below):

![edit_definition](https://user-images.githubusercontent.com/3840267/161601832-effd855a-fe1d-4b6c-8c41-c8d9c9b8ee64.png)

OR

2. Allow adding additional parent land cover classes in addition to the core 7 ones in the window below. If so, which default values (-1, 0, 1) should we assign on creation?

![edit_core_lc](https://user-images.githubusercontent.com/3840267/161602829-ff2e2a10-f21b-478f-a010-b4863bacef93.png)

OR

3. Both 1 and 2?

azvoleff:
It is the second:

> Allow adding additional parent land cover classes in addition to the core 7 ones

The tool to change the parent legend should be located on the trends.earth settings window, just above the "Trends.Earth login information" section (you will need to add a new section there).

[...]

PR #626, opened 2022-04-25
Land Cover Custom Legend

gkahiu:
Adds functionality defined in #580.


LDMP/calculate_lc.py
```diff
@@ -89,7 +89,7 @@ def calculate_on_GEE(self):
             "crs": self.aoi.get_crs_dst_wkt(),
             "crosses_180th": crosses_180th,
             'legend_nesting': LCLegendNesting.Schema().dump(
-                self.lc_setup_widget.aggregation_dialog.nesting
+                self.lc_setup_widget.aggregation_dialog._nesting
```
azvoleff, line 92: I think this should now be referencing `lc_setup.get_lc_nesting()`, no?

LDMP/calculate_ldn.py
```diff
@@ -302,7 +302,7 @@ def btn_calculate(self):
                 year_final,
                 'legend_nesting':
                 LCLegendNesting.Schema().dump(
-                    self.lc_setup_widget.aggregation_dialog.nesting
+                    self.lc_setup_widget.aggregation_dialog._nesting
```
azvoleff, line 305: Should now be referencing `lc_setup.get_lc_nesting()`?

LDMP/calculate_ldn.py
```diff
@@ -318,7 +318,7 @@ def btn_calculate(self):
                 .80,
                 'legend_nesting':
                 LCLegendNesting.Schema().dump(
-                    self.lc_setup_widget.aggregation_dialog.nesting
+                    self.lc_setup_widget.aggregation_dialog._nesting
```
azvoleff, line 321: Should now be referencing `lc_setup.get_lc_nesting()`?

LDMP/calculate_lc.py
gkahiu: I think it should revert to the previous one i.e. 
```
self.lc_setup_widget.aggregation_dialog.nesting
```

Unintended consequence of auto-refactoring as I did not modify this part.

LDMP/calculate_ldn.py
gkahiu: See previous comment...most likely have been as a result of auto-refactoring which was only in `lc_setup.py`

LDMP/calculate_ldn.py
gkahiu: Same...should be the `.nesting` without the underscore

azvoleff (commented):
Clicking the reset button on the transition matrix (below) resets the matrix to the original seven classes. The button I mean is below:

image

However the aggregation window remains set to the custom classes (if any) that a user has set. For example see below, where I'd removed the artificial class. Resetting the transition matrix restores it there, but not in the aggregation window:

image

This means the two are out of sync and referring to different class lists. Simplest way to fix this would be to make the reset button on the transition matrix restore the matrix to the less version saved in qsettings (if easy and saving to qsettings is done only on clicking ok on that dialog). Otherwise, we could just have the reset button on the transition matrix reset it to having all the transitions set to stable (if custom classes are set), or to the default transition matrix (current behavior, but make this happen only if no custom classes are set).


Also one other fix - there needs to be an option to nest classes under the "nodata" class in the parent legend on this screen:

image

gkahiu:
> This means the two are out of sync and referring to different class lists. Simplest way to fix this would be to make the reset button on the transition matrix restore the matrix to the less version saved in qsettings (if easy and saving to qsettings is done only on clicking ok on that dialog). Otherwise, we could just have the reset button on the transition matrix reset it to having all the transitions set to stable (if custom classes are set), or to the default transition matrix (current behavior, but make this happen only if no custom classes are set).

The Reset table button now restores the matrix in the settings so that its in sync with the land cover nesting.

> 
> Also one other fix - there needs to be an option to nest classes under the "nodata" class in the parent legend on this screen:
> 
> image

I have tried to incorporate this (but not included in this PR) but [this](https://github.com/ConservationInternational/trends.earth-schemas/blob/master/te_schemas/land_cover.py#L121) validation error is raised because LCLegend's `codes` function [here](https://github.com/ConservationInternational/trends.earth-schemas/blob/master/te_schemas/land_cover.py#L57) does not include `nodata`. I do not want to refactor this part as am not sure what implications it might have on dependencies. Let me know how to proceed.

gkahiu:
> 
> Also one other fix - there needs to be an option to nest classes under the "nodata" class in the parent legend on this screen:

Done...
image

But there is a still validation issue. Since `nesting.parent.nodata` and `nesting.child.nodata` will be constant, therefore we do not expect this mapping `"-32768": [-32768]` in the `nesting` attribute to change. However, if we add another land cover class, whose parent is "nodata", then the model will complain - due to [these](https://github.com/ConservationInternational/trends.earth-schemas/blob/master/te_schemas/land_cover.py#L112) checks - that there are duplicate codes since we now have a new entry under `nesting` that also refers to `-32768`. 

Grateful if you could advise on how to address the logical representation in `LCLegendNesting` object.

azvoleff:
> But there is still a validation issue. Since `nesting.parent.nodata` and `nesting.child.nodata` will be constant, therefore we do not expect this mapping `"-32768": [-32768]` in the `nesting` attribute to change. However, if we add another land cover class, whose parent is "nodata", then the model will complain - due to [these](https://github.com/ConservationInternational/trends.earth-schemas/blob/master/te_schemas/land_cover.py#L112) checks - that there are duplicate codes since we now have a new entry under `nesting` that also refers to `-32768`.
> 
> Grateful if you could advise on how to address the logical representation in `LCLegendNesting` object.

Those two checks should just catch repetition of classes within the parent or children class lists. So for example listing both `"-32768": [-32768]` and `"1": [-32768]` (or doing both `"-32768": [-32768]`, and `"-32768": [1]`) should each raise an exception. But listing multiple children under a single parent shouldn't be a problem - does `"-32768": [-32768, 1, 2, 3]` raise an exception? That is the situation we need to allow here - allowing a user to nest multiple child classes under the parent nodata class.

gkahiu:
The current implementation of custom classes adds each new class to `LCNesting`'s `parent` attribute, then attempts to (initially) map, through the `nesting` attribute, the corresponding child based on the original UNCCD parent class. So for instance, if I have _Test Class 1_, with a code 23 and based on _Cropland_ parent, then the nesting will look like:

```
"nesting":{..."23": [10, 11, 12, 20, 30, 30]...}
```

The `DlgCalculateLCSetAggregation` class, including `LCClassComboBox`, use the `parent` attribute of `LCNesting` for class mapping and selection. Hence, if I create another class with, say code 8, that inherits from nodata, then the nesting will appear as:

```
"nesting":{..."8": [32768], "-32768": [-32768]...}
```

As a result, I get the validation error because of multiple `-32768` children.

On the otherhand, if I include the custom classes under the `child` attribute of `LCNesting`, then these (custom classes) will be together with the _ESA CCI Land Cover_ classes and as such, in the current implementation of `DlgCalculateLCSetAggregation` and `LCClassComboBox`, will be listed under the _Input Class_ column. The 7 default UNCCD classes will remain constant in the `parent` attribute. You had mentioned we include them in the `child` attribute but the representation in the aggregation window will be different as explained above unless modify how `LCNesting` is represented in the `DlgCalculateLCSetAggregation` class.

Let me know otherwise am happy to discuss further.

azvoleff:
> The current implementation of custom classes adds each new class to `LCNesting`'s `parent` attribute, then attempts to (initially) map, through the `nesting` attribute, the corresponding child based on the original UNCCD parent class. So for instance, if I have _Test Class 1_, with a code 23 and based on _Cropland_ parent, then the nesting will look like:
> 
> ```
> "nesting":{..."23": [10, 11, 12, 20, 30, 30]...}
> ```
> 
> The `DlgCalculateLCSetAggregation` class, including `LCClassComboBox`, use the `parent` attribute of `LCNesting` for class mapping and selection. Hence, if I create another class with, say code 8, that inherits from nodata, then the nesting will appear as:
> 
> ```
> "nesting":{..."8": [32768], "-32768": [-32768]...}
> ```
> 
> As a result, I get the validation error because of multiple `-32768` children.

Given it is important that users be able to nest multiple custom different custom classes under the same parent class, I think the best solution will have to be dropping the attempt to auto-assign children to to the custom classes. While this would mean more work needed by the user to nest children under the custom classes, this would avoid the above validation issue. Also forcing users to have to think on which classes should nest under the custom ones they enter isn't necessarily a bad thing, so I think this is the best way forwards.
 
> On the otherhand, if I include the custom classes under the `child` attribute of `LCNesting`, then these (custom classes) will be together with the _ESA CCI Land Cover_ classes and as such, in the current implementation of `DlgCalculateLCSetAggregation` and `LCClassComboBox`, will be listed under the _Input Class_ column. The 7 default UNCCD classes will remain constant in the `parent` attribute. You had mentioned we include them in the `child` attribute but the representation in the aggregation window will be different as explained above unless modify how `LCNesting` is represented in the `DlgCalculateLCSetAggregation` class.
> 
> Let me know otherwise am happy to discuss further.

Agreed, we don't want to go this route - custom classes need to show up within the parent classes so that the ESA-CCI can be aggregated within them.

gkahiu:
Thanks @azvoleff!

Based on this, I have incorporated the option of specifying ESA CCI child classes when adding or updating a parent land cover class so that the user can define a complete land cover nesting. Please test and let me know.

gkahiu:
Hi @azvoleff , any feedback on this PR?

azvoleff:
Thanks for this @gkahiu. This is almost there. A few more comments from me are below.

Also, @gabrieldaldegan - I've created a [specific zipfile for this branch](https://s3.us-east-1.amazonaws.com/trends.earth/sharing/LDMP_land_cover_branch.zip) so you can test this. Note this is a different zipfile from the latest development version zipfile that is linked to from the main webpage (I didn't want to put this version up there until we have it fully tested).

- The main settings window within which this is embedded is coming up with a very small window size (so there are both horizontal and vertical scrollbars when that window opens) can we have it default to coming up a bit larger so that at least the horizontal scrollbar isn't needed?

image

- Can we make the "Custom Land Cover Classes" section of the settings window default to a slightly larger height? Right now you need to scroll even when just the default 7 classes are being used:
image

- Increasing the height of the "Custom Land Cover Classes" section of the settings window may address this as well, but right now you can't see even the done button without scrolling after you click to edit a class:

image

- There needs to be a way to assign custom classes that have no children nested underneath them. This is necessary as a common use case will be users who are interested only in adding additional cover classes so that they can represent a specific legend they use in their country. In many cases these users are using the import functions in Trends.Earth to load their own data - these users therefore have no interest in assigning nesting for the ESA cover classes ("Tree cover, needle", "Tree cover, broadleaf", etc.) as the ESA class information is only needed for users that use the default datasets in Trends.Earth.

- I am a bit torn on the new "Child Class Mapping" screen. I like it as it is cleaner than the old dropdown layout, but I worry it might be more of a hassle to use and require more clicking to make reassignments of child classes among different parents. @gabrieldaldegan would likely have more thoughts on this.

New screen:
image

Old screen:
image

gkahiu:
> Thanks for this @gkahiu. This is almost there. A few more comments from me are below.
> 
> Also, @gabrieldaldegan - I've created a [specific zipfile for this branch](https://s3.us-east-1.amazonaws.com/trends.earth/sharing/LDMP_land_cover_branch.zip) so you can test this. Note this is a different zipfile from the latest development version zipfile that is linked to from the main webpage (I didn't want to put this version up there until we have it fully tested).
> 
> * The main settings window within which this is embedded is coming up with a very small window size (so there are both horizontal and vertical scrollbars when that window opens) can we have it default to coming up a bit larger so that at least the horizontal scrollbar isn't needed?
> 
> image
> 
> * Can we make the "Custom Land Cover Classes" section of the settings window default to a slightly larger height? Right now you need to scroll even when just the default 7 classes are being used:
>   image
> * Increasing the height of the "Custom Land Cover Classes" section of the settings window may address this as well, but right now you can't see even the done button without scrolling after you click to edit a class:
> 
> image
> 

Alright, will adjust the sizes accordingly.

> * There needs to be a way to assign custom classes that have no children nested underneath them. This is necessary as a common use case will be users who are interested only in adding additional cover classes so that they can represent a specific legend they use in their country. In many cases these users are using the import functions in Trends.Earth to load their own data - these users therefore have no interest in assigning nesting for the ESA cover classes ("Tree cover, needle", "Tree cover, broadleaf", etc.) as the ESA class information is only needed for users that use the default datasets in Trends.Earth.

If there are no children assigned then `LCLegendNesting` will raise validation errors. The alternative is to go back to the option of auto-assigning children based on the parent.

> * I am a bit torn on the new "Child Class Mapping" screen. I like it as it is cleaner than the old dropdown layout, but I worry it might be more of a hassle to use and require more clicking to make reassignments of child classes among different parents. @gabrieldaldegan would likely have more thoughts on this.
> 
> New screen: image
> 
> Old screen: image

I actually preferred using the old screen as it is familiar, however the dialog requires a complete `LCLegendNesting` i.e. one where the parents and corresponding children have been defined. However, since the classes are incomplete (without children) at the time of creation then this would not have worked also due to validation errors hence the reason to add the option of defining children when the parent is created. Am happy to brainstorm further on this.

azvoleff:
> > * There needs to be a way to assign custom classes that have no children nested underneath them. This is necessary as a common use case will be users who are interested only in adding additional cover classes so that they can represent a specific legend they use in their country. In many cases these users are using the import functions in Trends.Earth to load their own data - these users therefore have no interest in assigning nesting for the ESA cover classes ("Tree cover, needle", "Tree cover, broadleaf", etc.) as the ESA class information is only needed for users that use the default datasets in Trends.Earth.
> 
> If there are no children assigned then `LCLegendNesting` will raise validation errors. The alternative is to go back to the option of auto-assigning children based on the parent.

We have two different nestings here:

1. custom legend (children) under 7 IPCC classes (parent)
2. ESA land cover classes (children) under custom legend (parent)

There is no need to set the ESA legend nesting unless a user wants to use the ESA classes. The ESA classes are only used in three cases:

1. "Sub-indicators for SDG 15.3.1" tool
2. "Land cover change" tool when executed remotely
3. "Soil organic carbon" tool when executed remotely

So what I'd suggest is that the current workflow modify only the nesting of the custom legend under 7 IPCC classes. The nesting of the ESA land cover classes under that custom legend can be set if and only if a user tries to run one of the above three tools. In that case we can try to auto-assign children based on the parent where possible, and then force the user to manually set any that can't be auto-assigned. This approach would ensure that we can always have children assigned for each parent class, but only require users to go through this process if they actually need to use the ESA legend. Would this make sense?

> > * I am a bit torn on the new "Child Class Mapping" screen. I like it as it is cleaner than the old dropdown layout, but I worry it might be more of a hassle to use and require more clicking to make reassignments of child classes among different parents. @gabrieldaldegan would likely have more thoughts on this.
> > 
> I actually preferred using the old screen as it is familiar, however the dialog requires a complete `LCLegendNesting` i.e. one where the parents and corresponding children have been defined. However, since the classes are incomplete (without children) at the time of creation then this would not have worked also due to validation errors hence the reason to add the option of defining children when the parent is created. Am happy to brainstorm further on this.

See above - we could stick with the only screen if we then force users to use it IFF they try to use a tool that needs to use the ESA legend nesting.

gkahiu:
Thanks @azvoleff. Some follow-up questions:

1. By referring to IPCC, do you mean the UNCCD parent classes in `land_cover_nesting_unccd_esa.json`?
2. Am assuming the transition matrix widget will always show the custom legend regardless of the algorithm. Is this correct?

image

> There is no need to set the ESA legend nesting unless a user wants to use the ESA classes. 
3. With regard to the comment above, will the user have an option to select which the legend nesting to use in the algorithm window or will this be set at the configuration level say by using a flag under `additional_configuration` in `scripts.json` to indicate which legend nesting to use so as to avoid hardcoding using the algorithm name?
4. On first time use, will the user be required to define a custom legend or do we provide a default option? You had previously mentioned we nest the UNCCD classes under themselves, is this still the case?

azvoleff:
> Thanks @azvoleff. Some follow-up questions:
> 
> 1. By referring to IPCC, do you mean the UNCCD parent classes in `land_cover_nesting_unccd_esa.json`?

Yes, thats right.

> 2. Am assuming the transition matrix widget will always show the custom legend regardless of the algorithm. Is this correct?

Yes, correct
 
> image
> 
> > There is no need to set the ESA legend nesting unless a user wants to use the ESA classes.
> 
> 3. With regard to the comment above, will the user have an option to select which the legend nesting to use in the algorithm window or will this be set at the configuration level say by using a flag under `additional_configuration` in `scripts.json` to indicate which legend nesting to use so as to avoid hardcoding using the algorithm name?

Probably easier to chat on this, but what I was envisioning is that the ESA classes would be nested under the custom classes that a user set, so there wouldn't be multiple different nesting options. We could add a flag under `additional_configuration` to indicate when an algorithm needs the ESA nesting to be set - is that what you are asking?

> 4. On first time use, will the user be required to define a custom legend or do we provide a default option? **You had previously mentioned we nest the UNCCD classes under themselves, is this still the case?**

Yes, , we should nest the UNCCD classes under themselves by default.

gkahiu:
A few notes on the update:

1. On defining custom LC classes, two LC nestings (i.e. IPCC and ESA) are created and saved in the settings. The `LandCoverSetupRemoteExecutionWidget` provides an option to specify which nesting to use between the two;
2. For the IPCC nesting, if there are custom classes whose parents are 'nodata', then the `nodata` attribute of the child legend in the nesting is set to `None` in order to avoid any validation errors;
3. Similarly for the ESA nesting:
    - If there are multiple custom classes that inherit from the same parent (e.g. Cropland), then the mapping to the child ESA classes is only applied to the first custom class during the auto-assignment process. The mapping to subsequent (custom) classes is not defined (i.e. empty list in the nesting attribute) but can be manually specified in the LC aggregation dialog;
    - Any ESA classes that are not assigned a parent custom class are removed from the child nesting in order to avoid the validation errors.
4. I manually configured the "Sub-indicators for SDG 15.3.1", "Land cover change (Remote)" and "Soil organic carbon (Remote)" tools to use the ESA nesting rather than specifying this in the script configuration as discussed earlier. I realized the initialization of `LandCoverSetupRemoteExecutionWidget` is different across these tools hence there was not much of an advantage in doing it through the script configuration.

I hope these changes match the expected workflow.

Issue #579, opened 2022-03-22
Need threadsafe updates crash resolved


timlinux:
@azvoleff please share some diagnostics.

azvoleff:
And one more trace, which came up during a background update of the trends.earth datasets window. Unforrtunately not too much in this one:


## User Feedback

QGIS open to Trends.Earth plugin page. Update of dataset status proceeding in background.

## Report Details

**Python Stack Trace**
```
Windows fatal exception: access violation

Current thread 0x00004484 (most recent call first):
<no Python frame>
```


**Stack Trace**
No stack trace is available.


**QGIS Info**
QGIS Version: 3.24.0-Tisler
QGIS code revision: 6b44a42058
Compiled against Qt: 5.15.2
Running against Qt: 5.15.2
Compiled against GDAL: 3.4.1
Running against GDAL: 3.4.1



**System Info**
CPU Type: x86_64
Kernel Type: winnt
Kernel Version: 10.0.19044

gdaldegan:
got this one when moving QGIS from laptop screen to external screen

Stack Trace

QPainter::drawPixmap :
PyInit_QtGui :
PyArg_ParseTuple_SizeT :
PyEval_EvalFrameDefault :
PyObject_GC_Del :
PyFunction_Vectorcall :
PyFloat_FromDouble :
PyVectorcall_Call :
PyObject_Call :
PyInit_QtWidgets :
QTreeView::drawRow :
PyInit_QtWidgets :
QTreeView::drawTree :
QTreeView::paintEvent :
PyInit_QtWidgets :
QWidget::event :
QFrame::event :
QAbstractItemView::viewportEvent :
PyInit_QtWidgets :
QCoreApplicationPrivate::sendThroughObjectEventFilters :
QApplicationPrivate::notify_helper :
QApplication::notify :
QgsApplication::notify :
QCoreApplication::notifyInternal2 :
QWidgetPrivate::drawWidget :
QWidgetPrivate::paintSiblingsRecursive :
QWidgetPrivate::drawWidget :
QWidgetPrivate::paintSiblingsRecursive :
QWidgetPrivate::drawWidget :
QWidgetPrivate::paintSiblingsRecursive :
QWidgetPrivate::drawWidget :
QWidgetPrivate::paintSiblingsRecursive :
QWidgetPrivate::paintSiblingsRecursive :
QWidgetPrivate::drawWidget :
QWidgetPrivate::paintSiblingsRecursive :
[...]

PR #649, opened 2022-05-16
Fix thread crash

gkahiu:
Hi Alex,

Unfortunately I still cannot reproduce the crash issue hence will try to isolate possible causes. What we know is that it is caused by local jobs calling `submit_local_job_as_qgstask` in the `JobManager`, which might mean two things:

1. Calls to `LocalJobTask` might be leading - in some way - to references to objects in the main UI thread (maybe the item model or widget); OR,
2. Calls to underlying algorithms might be have some errors maybe due to shared file handles.

For this initial PR, the call to `submit_local_job_as_qgstask` is just a proxy call to `submit_local_job` to see if the crash still persists if we eliminate the task.

Kindly test and let me know.


azvoleff:
Thanks John. I've tested this and it seems to be working thus far on my side. Note I made a few smell changes - just reverting a few inadvertent changes to version numbers for the scripts.

@gabrieldaldegan this is in the latest zipfile. Can you post to #579 if you still see crashes after this change?

PR #657, opened 2022-06-16
NDVI Query Fix

gkahiu:
Minor fix that ensures query values have been extracted before resetting the widgets.


azvoleff:
Thanks. @gkahiu, and just FYI - right now the timeseries tasks will be failing for a bit while I make API updates to that endpoint

PR #686, opened 2022-08-03
Fix loading DlgTimeseries.ui on Linux

mentaljam:
Hi,

Thanks for all your work on Trends.Earth! I've faced an error loading the plugin on Linux because it uses case-sensitive paths. Here is a fix.


azvoleff:
Thanks! Had missed this one.

Issue #694, opened 2022-10-10
Refractoring to continuous integration tests


vermeulendivan:
Continuous Integration testing provides a critical tool to ensure that regressions (breakages with the introduction of new changes) do not occur in the code base. In this work package with will migrate the integration testing suite to use the GitHub Actions system for automating tests. The goal will be to automatically prevent the merging of any changes if the test framework does not pass. In addition, we will use any remaining hours in this deliverable to make other improvements to the testing framework such as expanding the number of tests, automatic creation of build artifacts (that can be downloaded and tested in a user’s local QGIS session) and ensuring that the test framework uses latest current best practice for the QGIS project.
Steps:

1. Samweli to perform a quick review of which framework to use
2. Any new functionality entering into the codebase should be accompanied by test
3. Retroactively add tests for key components like land cover class functionality
4. GUI testing is also needed so we need to review options e.g. Victor Olaya’s test tool versus Qtest framework, versus other test frameworks.

Samweli:
Related to https://github.com/ConservationInternational/trends.earth/issues/664

Issue #664, opened 2022-07-02
Update tests


azvoleff:
- [ ] fix to run on github actions
- [ ] add tests for code accessing the API
- [ ] add tests for each algorithm (including remote and localexecution modes)

PR #704, opened 2022-10-27
CI workflow for tests 

Samweli:
Contains fix for https://github.com/ConservationInternational/trends.earth/issues/664 and https://github.com/ConservationInternational/trends.earth/issues/694
Adds GitHub action workflow for running tests in the CI during code pushes and pull requests.
The test workflow runs against QGIS 3.22 LTR, 3.26 and master branch images.


admin.py
```diff
@@ -0,0 +1,301 @@
...
+    :type qgis_profile: str
+    """
+    context.obj = {
+        "verbose": verbose,
+        "qgis_profile": qgis_profile,
+    }
+
+
+@app.command()
+def generate_zip(
```
azvoleff, line 73: The plugin uses a number of dependencies for which code to generate a zipfile that packages them with it is now located in the invoke tasks in `tasks.py`. We'll either need to make this function also handle packaging those dependencies, or else borrow from the various development-related tasks that we already have code for in `tasks.py` to make sure this is all handled correctly.

Samweli: @azvoleff sure, I have removed the `admin.py` and will open another PR for it. I'm keeping this PR focused on only running the tests on the CI

PR #705, opened 2022-10-27
Settings migration to QGIS options, and other changes

### LINKED ISSUE #667: Move trends.earth settings into main QGIS settings window
Currently settings for Trends.Earth are accessed from the plugin toolbar:

image

This brings up what has become a very cramped window:

image

We should move these settings into the main QGIS options screen:

image

We do need to keep the functionality to change regions and the land cover legend easily accessible though, so we'll need a way for a user to click on the "Change region" button in the various tools (and similarly for land cover legend) and directly bring up that page in the main QGIS options window.

### LINKED ISSUE #693: Migrate settings to QGIS options
The QGIS project provides a framework for settings management and it is
recommended to utilise this framework in plugins too. Any plugin may add a new tab to
the settings dialog as shown in the example below:

image

In this work deliverable we will migrate the existing options (e.g. when you click on the ‘wrench’ icon)
into the QGIS Options dialog in a tab called trends.earth. In addition, we propose to add some logic so
that it is clear for users running the plugin for the first time (i.e. with no account yet configured) that
they should go to the settings dialog to be onboarded.

image

Custom land cover classes and the current region settings are accessible from multiple places in the
UI. We need these still to work but open the central settings dialog with the correct tab open etc.
Also note the screenshots above are out of date so consult the latest version of the plugin.

### LINKED ISSUE #701: Custom landcover classes size should be increased
Currently if the number of classes added to the custom land cover classes in settings is increased to a large amount, it does not fit in the settings panel nicely. The best approach will be to increase the actual size of the table for the landcover classes to increase as the number of classes increases.
NOTE: Class count should not exceed 38.

image

### LINKED ISSUE #702: Move the Trends.Earth login info
Move the Trends.Earth login information section (highlighted in red) in the settings to the top of the Trends.Earth settings tab.

image

vermeulendivan:
@azvoleff @timlinux Here is the changes thusfar, and changes made after the meeting. Minor changes not discribed in this PR.

@azvoleff I first updated the branch on my side, so there should be no code clashes.

fixes #702 fixes #701 fixes #693 fixes #667 

Trends.Earth settings has been moved to QGIS option panel:
image

Login info moved to the top, and both login info and ROI has been changed to collapsible boxes/containers
image

Landcover classes can extent to a larger height. Minimum height is 4 rows.
image
image
image

Replaced the Python raise error for Drought vulnerability summary table with a QGIS message bar message. This is an example/suggestion for replacing other Python raise error cases (there is a lot).
image


azvoleff:
Thanks @vermeulendivan. Can you please format with black before committing? Looks like that check is failing.

LDMP/select_dataset.py
```diff
@@ -62,16 +62,11 @@ def lc_band(self):
     def soil_band(self):
         return self.combo_soil.get_current_band()
 
-    def sdg_band(self):
```
azvoleff, line 65: Why are you removing this function? It is still used in `calculate_ldn.py` and `jobs/mvc.py`.

azvoleff (commented):
Can you also move the changes to `areaofinterest.py` and `calculate.py` to a separate pull request? Some of the classes that subclass `DlgCalculateBase` I believe check for and catch exceptions like that currently raised if a buffer isn't selected, so that users can fix the issue without the dialog they are using closing (which would lead to them losing their entries in it). So we'll need to make sure that changing that from an exception to a notification won't mess up any of the windows that are subclasses of `DlgCalculateBase`

azvoleff:
@vermeulendivan we still need to figure out a better solution to the changing size of the custom land cover box as the legend size changes. As an example: if you click "Restore" and select the 37 class UNCCD legend:

image

You then get a very long list of classes in that box:

image

If I then scroll down and click a class near the bottom of the list, say the "Artificial" class, it loads the Land Cover Class Editor widget to edit that class. But the widget isn't visible because it is much farther up in the scrolling area:

image

This is confusing for the user as they need to immediately scroll up just after clicking a class in order to actually edit it. So maybe we need a shorter maximum height for the land cover class widget (with an added scroll bar), or some other way to ensure that the Land Cover Class Editor is always visible immediately after clicking to edit a class.

azvoleff:
One other usability fix: right now clicking a class in the Custom Land Cover Classes list brings up the editor, and there is no way to select and delete multiple classes quickly. Can we make it so:

- a double-click or similar brings up the editor
- enable selection of multiple classes with the mouse (click+drag and/or shift) and enable deletion of a selected class with delete key

LDMP/select_dataset.py
vermeulendivan: @azvoleff Removed it because its a duplicate function (see line 53). See https://github.com/ConservationInternational/trends.earth/blob/f88fae8601ea281db3f5cd88da6d86ceaa56b9f1/LDMP/select_dataset.py#L53. So the function is still there, and it will cause no issues.

azvoleff: Whoops - sorry, I read through too fast.

azvoleff:
Thanks. All looks good except the resizing of the add class widget. It still is a little funky when the class list is long. For example if I rest to 37 classes, then scroll, my settings window looks like this:

qgis-bin_HhPTcDSFxb

If I click on that class then my screen looks like this:

qgis-bin_w6GO2XVF8A

In other words the add class widget isn't visible at all unless I scroll up to it, and even then it doesn't fit within the space it is given (at least on my screen), so I wind up with it having a scrollbar:

image

Maybe the best is if we go back to a fixed size window large enough to display 10-12 classes. That should fit most of the cases, and allow the larger ones to show without too much scrolling / without the weird jumps in size we get right now.

vermeulendivan:
@azvoleff The settings code is ready to be merged. There were conflicts, which I have already sorted out.

azvoleff:
Thanks! Looks good to me!

PR #730, opened 2022-11-23
Fix current failing CI tests

Samweli:
These changes contain solution for current repository CI tests fails, by introducing the following updates
- Removed all Qt imports that used `PyQt5` directly and now using `qgis.PyQt` as recommended by the QGIS development guidelines.
- Updates to the missing `QgsPanelWidget` now using `QWidget`
- Only installing the necessary package needed to run tests  stored from the new `requirements-testing.txt` file.
- Using a new tests discovering script `test_suite.py`


Samweli:
Hi @azvoleff, do you know where I can find data for files called `Nepal_Narayani_15_3_1_TE_One_Step.tif`  and `Surinam_Paramaribo_11_3_1_Urban_10km.tif` used in these tests respectively https://github.com/ConservationInternational/trends.earth/blob/fb0d402a20716b4cc28dfac0a873215aa03f8f3c/LDMP/test/integration/test_calculate_ldn.py#L13 and https://github.com/ConservationInternational/trends.earth/blob/fb0d402a20716b4cc28dfac0a873215aa03f8f3c/LDMP/test/integration/test_calculate_urban.py#L25

Samweli:
ping @azvoleff

requirements-testing.txt
```diff
@@ -0,0 +1,6 @@
+boto3
+black[jupyter]
+plotly
+pyqtgraph
+git+https://github.com/ConservationInternational/trends.earth-schemas.git@v2.1.2
```
azvoleff, line 5: This is a requirement of the plugin itself (not just of the tests), so it would be better to manage this through the regular plugin dependencies than separately here (otherwise this is just another place to keep the correct version number in sync with that specified in the regular requirements.txt file). [This function](https://github.com/ConservationInternational/trends.earth/blob/main/tasks.py#L658) should be used to install those dependencies when the plugin is being setup on the server for the CI tests.

requirements-testing.txt
```diff
@@ -0,0 +1,6 @@
...
+git+https://github.com/ConservationInternational/trends.earth-algorithms.git@v2.1.2
```
azvoleff, line 6: Same comment as above - need to use the plugin_setup invoke task to pull in the plugin dependencies to ensure that the CI sever isn't using a different setup from that shipped with the plugin.

docker/trends-earth-test-pre-scripts.sh
```diff
@@ -6,5 +6,5 @@ qgis_setup.sh LDMP
 rm -rf  /root/.local/share/QGIS/QGIS3/profiles/default/python/plugins/trends.earth
 ln -sf /tests_directory/LDMP /root/.local/share/QGIS/QGIS3/profiles/default/python/plugins/trends.earth
 
-pip3 install -r /tests_directory/requirements.txt
```
azvoleff, line 9: Should be using this line, not the next (or better yet call the invoke task to setup the plugin dependencies directly), so that we test with the correct packages shipped with the plugin.

azvoleff (changes requested):
See in-line comments - we should be using the existing plugin-setup invoke task so that we ensure we test with the package dependencies shipped with the plugin (there are also invoke tasks to change the plugin versions between dev and release, etc., so we'd need to write more code to handle those tasks if we don't reuse that script for the CI server)

azvoleff:
> Hi @azvoleff, do you know where I can find data for files called `Nepal_Narayani_15_3_1_TE_One_Step.tif` and `Surinam_Paramaribo_11_3_1_Urban_10km.tif` used in these tests respectively
> 
> https://github.com/ConservationInternational/trends.earth/blob/fb0d402a20716b4cc28dfac0a873215aa03f8f3c/LDMP/test/integration/test_calculate_ldn.py#L13
> 
> and
> https://github.com/ConservationInternational/trends.earth/blob/fb0d402a20716b4cc28dfac0a873215aa03f8f3c/LDMP/test/integration/test_calculate_urban.py#L25
> 
> They are currently not available on the test folder. Other files that are missing and causing the PR tests runs to fail are `trends.earth_test_user_credentials.json` and `trends.earth_admin_user_credentials.json` referenced in
> 
> https://github.com/ConservationInternational/trends.earth/blob/3420705fce4310a9fbaea961e7a394a7a70520d3/LDMP/test/__init__.py#L15

Yes - I'll add these when I get back from travel (I get back on Tues next week). Likely I'll need to update those tests as well for the most recent algorithm versions, so can handle that as well once I get back.

Samweli:
> I've created [another branch](https://github.com/ConservationInternational/trends.earth/tree/linux-fixes) with the various linux fixes from this pull request so we can separate them out there (I'd rather not commit a change to main right now that swaps out `QgsPanelWidget` for `QWidget` given the problems it creates).

I have opened a PR that provides a fix only for the `QgsPanelWidget` issue here https://github.com/ConservationInternational/trends.earth/pull/736

azvoleff:
Thanks for all the work on this @Samweli! Have tested on my side and with the various `QgsPanelWidget`, fixes, etc., all looks good. Just a few test errors to be fixed (which is on my todo list - not yours). Will merge.

PR #735, opened 2022-11-30
Fix exception on newer python versions

nyalldawson:
Setting a QProgressBar to a float value raises an exception on Python 3.9 +


azvoleff:
Thanks!

Issue #738, opened 2022-12-01
Plugin error when algorithms results don't return anything


Samweli:
When running algorithms remotely its currently possible for the returned results to be `None` hence accessing any attributes from the results will cause the below error.
![image](https://user-images.githubusercontent.com/2663775/205013284-6b977f3e-535e-4387-a03b-f69e9b3b4a6c.png)

azvoleff:
Fixed by [d29ccb6](https://github.com/ConservationInternational/trends.earth/commit/d29ccb693639234503cbbec374faf87ecb46f1d2).

PR #739, opened 2022-12-01
Fix error in setting job extent

Samweli:
Fix for https://github.com/ConservationInternational/trends.earth/issues/738
These changes ensure that jobs are only set with results extent when the results are avaialable, this  also prevents accessing `None` results which can lead to `AttributeError`.


azvoleff:
Thanks @Samweli!

PR #742, opened 2022-12-05
Add a mock server for tests

Samweli:
This PR works on the following changes that are related to the tests.

- Updates on the `api` module now using a `APIClient` class to store and interface all the API related functionality this makes the module testable.
- Removal of unused methods in `calculate.py`
- Added a new mock API server for running tests local and on CI
- Include initial test for the `calculate_ldn` alg.


test/mock/data/sample_raw_job.json
```diff
@@ -0,0 +1,38 @@
...
+      "prod_state_year_tg_start": 2013,
+      "prod_traj_year_final": 2015,
+      "prod_traj_year_initial": 2001,
+      "task_name": "",
+      "task_notes": "",
+      "trajectory_method": "ndvi_trend"
+    },
+    "progress": 0,
+    "results": {},
+    "script_id": "47b99c3d-f5b5-4df3-81e7-324a17a6e147",
```
azvoleff, line 33: script_id needs to be kept in sync with latest value taken from [`scripts.json`](https://github.com/ConservationInternational/trends.earth/blob/main/LDMP/data/scripts.json#L132). These are kept in sync by [an invoke task](https://github.com/ConservationInternational/trends.earth/blob/main/tasks.py#L502). Script IDs change when there are upstream changes in trends.earth-algorithms or trends.earth-schemas, or when the GEE code used on the API changes

test/test_calculate_ldn.py
```diff
@@ -0,0 +1,89 @@
...
+        cfg.setConfig("username", "test")
+        cfg.setConfig("password", "test")
+        am.storeAuthenticationConfig(cfg)
+
+        QtCore.QSettings().setValue(f"trends_earth/{auth_setup.key}", cfg.id())
+
+        return cfg.id()
+
+    def test_job_submission(self):
+        ldn_one_step_script_id = "47b99c3d-f5b5-4df3-81e7-324a17a6e147"
```
azvoleff, line 71: Better to use script name than id (as id changes with script versions, while script name can be used in lieu of ID and stays constant). If you use a script uuid you'll need to add code to an invoke task to keep these script_ids in sync as they change across plugin versions

azvoleff (commented):
Thanks for this @Samweli. In general this looks great. I added a few inline comments regarding how script IDs are handled as this will be important for maintainability. But if you can rebase this relative to develop and submit again I'll merge it to the develop branch so you can continue work on the other issues as we discussed today

azvoleff:
@Samweli - we discussed me merging this so you could move on with other things. I don't want to hold you back, but this one is still marked as draft so I'm assuming it isn't ready to go. Let me know when this is ready to merge and I'll do so (also if you can resubmit relative to the `develop` branch).

Samweli:
> @Samweli - we discussed me merging this so you could move on with other things. I don't want to hold you back, but this one is still marked as draft so I'm assuming it isn't ready to go. Let me know when this is ready to merge and I'll do so (also if you can resubmit relative to the `develop` branch).

Hi @azvoleff thanks, I have opened another PR based on the `develop` branch here https://github.com/ConservationInternational/trends.earth/pull/745 and it is ready for review and merge.

PR #745, opened 2022-12-08
Add a mock server for tests [develop]

Samweli:
Based on https://github.com/ConservationInternational/trends.earth/pull/742, 

Adds the following changes that are related to the tests.

- Updates on the `api` module now using a `APIClient` class to store and interface all the API related functionality this makes the module testable.
- Removal of unused methods in `calculate.py`
- Added a new mock API server for running tests local and on CI
- Include initial test for the `calculate_ldn` alg.


azvoleff:
I get an immediate crash if I try to run this branch in QGIS, so I'm hesitant to merge at this point as I it would complicate testing any other PRs relative to develop. 

I don't get a crash report when it crashes, but I do see this in the log pane (looks like a missing method):

image

Are you able to run this on your machine?

Samweli:
> I don't get a crash report when it crashes, but I do see this in the log pane before QGIS closes (looks like a missing method):

@azvoleff same thing happened to me, I have made a fix for it here https://github.com/ConservationInternational/trends.earth/pull/745/commits/60f5ae6ce64e53bc3ce04da860d56e74b47c095a

LDMP/api.py
```diff
@@ -126,324 +144,339 @@ def finished(self, result):
...
-            )
-            """
             ret = None
-    else:
-        ret = None
 
-    return ret
+        return ret
 
+    def get_header(self, url):
```
azvoleff, line 360: This function doesn't actually use the API, so would make more sense to handle this in a different class (it is for getting the header of any generic URL, needed for checking file size, etc.)

azvoleff:
Still hesitant to merge this one @Samweli. It seems to break task submission - when you try to submit a remote task it appears nothing happens (the job doesn't show up on our server nor in the plugin, and there are no error messages shown to the user). It seems to be an issue with getting the token.

Samweli:
> when you try to submit a remote task it appears nothing happens

@azvoleff this was indeed an issues in getting the `token`, I have fixed and tested it see the below screenshot
image

azvoleff:
Thanks - looks good~

PR #746, opened 2022-12-08
Dev landcover matrix

### LINKED ISSUE #725: Make size of land cover degradation matrix larger when a large number of classes is chosen
When a large number of land cover classes are chosen, the window to setup the land cover degradation matrix should increase in size so that more rows of the matrix are shown (below is from the SDG 15.3.1 tool, but same goes for the land cover tool as well):

image

vermeulendivan:
fixes #725 

The degradation landcover table will now increase in size as the number of rows increases - this will avoid an additional scroll bar for the table:
image

If there are 5 or less rows in the table, the row height will be set to 40. The value 40 is used because its the lowest value that can be used, due to the minimum value based on the font used for the table cells.
image

### DIFF
--- a/LDMP/conf.py
+++ b/LDMP/conf.py
@@ -29,8 +29,10 @@ OPTIONS_ICON = "trends_earth_logo_square_32x32.ico"
 OPTIONS_TITLE = "Trends.Earth"
 
 # Degradation Landcover matrix size settings
-RESIZE_NUM_ROWS = 5  # If there are equal to less than this value rows than the row height will be decreased
-MINIMUM_ROW_HEIGHT = 40  # This is the minimum row height value. 40 is the minimum at the moment, based on the font
+# If there are equal to less than this value rows than the row height will be decreased
+RESIZE_NUM_ROWS = 5
+# This is the minimum row height value. 40 is the cell font minimum
+MINIMUM_ROW_HEIGHT = 40
 
 class AreaSetting(enum.Enum):
     COUNTRY_REGION = "country_region"
--- a/LDMP/conf.py
+++ b/LDMP/conf.py
@@ -34,6 +34,7 @@ RESIZE_NUM_ROWS = 5
 # This is the minimum row height value. 40 is the cell font minimum
 MINIMUM_ROW_HEIGHT = 40
 
+
 class AreaSetting(enum.Enum):
     COUNTRY_REGION = "country_region"
     COUNTRY_CITY = "country_city"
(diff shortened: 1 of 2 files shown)


vermeulendivan:
@azvoleff @Samweli The tests which fails here are not related to any changes I've made, and looking it the failures does not refer to changes in this PR. I'm not sure why these errors occurs.

azvoleff:
A few more changes after testing. Right now it doesn't seem to scroll correctly horizontally. Can you please test it on your end to make sure it is usable with the 37 class matrix that you can access from the "restore" button in the land cover class setup pane? If I use that 37 class legend I get a window like this if I expand it:

image

So I can see all the classes horizontally if I make the window huge, but if the window shrinks down horizontally (remembering many users will have low resolution screens on older laptops), then the last few columns disappear even when I use the horizontal scrollbar that appears:

image

Note I've scrolled all the way to the right but still can't see the last few classes - they are cut off just after the "Shrub or herbaceous" class.

There is also one other usability issue: if a user scrolls down on a large matrix they lose the column headers so they have no idea what each column is. For example see the below:

image

There isn't any easy way to modify these classes since you can't match columns to their class names for the rows at the end of the list except by scrolling up and down constantly. Two ways I can see to try to fix this:

- Have the mouseover text show something like "Row name, Column name" for each cell when you mouse over it
- Freeze the column header on the top so it is always visible even when you scroll vertically. This would require having the ability to scroll the land cover pane independently of the rest of that window

vermeulendivan:
@azvoleff Hi cannot replicate the issue you mention above on the last class(es) not showing horizontally. Its fine on my side:
image
This test is on a laptop and I even made the calculate dialog very small. Take note that there are a horizonal scroll bar for the table itself which you need to move.
Do I need to make the table bigger horizontally as well, so that there are only one horizonal scroll bar (e.g. one for the entire calc dialog, and none for the table)? I'm assuming yes.

vermeulendivan:
@azvoleff No worries, got it working to also horizonally increase the table size - so now there are only on scroll bar. Apologies, should have taken the horizonal size into account.
Will look at the column headers show that the user can see it when scrolling around.

vermeulendivan:
@azvoleff I've added a commit which will auto resize the horizonal axis of the table as well, so horizonally there will also now only be one scroll bar (the main calc dialog). Had a look at adding freezing for the headers, but from online research there is no easy way to implement this.

azvoleff:
Thanks - this looks good. I just added a tooltip to it. Good to go!

PR #749, opened 2022-12-12
Offline mode support

### LINKED ISSUE #697: Offline mode for the plugin
Add to the settings an offline mode for when a user needs to work offline (e.g. no internet connection close available). Add to the QGIS plugin settings panel:
image

Should cover the following:
- Hide/disable function or algorithms which is not available during offline mode
- Tell the user if offline mode is active
- R&D: Check if its possible to add a bandwidth speed tester for slow internet connections

vermeulendivan:
fixes #697 

Adds offline mode support to the plugin. The user can enable the offline mode by going to the Trends.Earth options, in the Advanced section.
image

Login info disabled
image

Remote algorithms are disabled:
image

- The dock title will change to "Trends.Earth (offline mode)".
- Periodic calls to the server (e.g. refresh data list) is disabled.
- Download dataset button disabled
- Remote processes will no longer be displayed in the list. Disabling offline mode and clicking refresh will show them again

### DIFF
--- a/LDMP/algorithms/mvc.py
+++ b/LDMP/algorithms/mvc.py
@@ -7,6 +7,7 @@ from qgis.PyQt import QtGui
 from qgis.PyQt import QtWidgets
 from qgis.PyQt import uic
 from te_schemas.algorithms import AlgorithmRunMode
+from ..conf import settings_manager, Setting
 
 from . import models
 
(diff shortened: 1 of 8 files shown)


azvoleff:
This looks great! One minor issue I've noticed @vermeulendivan - when enabling/disabling offline mode some of the tools don't get enabled/disabled correctly. For example see below after I turn off offline mode,  you should now be able to run the drought tool remotely but it is still disabled. I've seen the same with the SDG tool as well where it doesn't seem to sync correctly as the mode changes.

azvoleff:
@vermeulendivan is this ready to merge? Were you able to address the issue with the "execute remotely", etc., buttons getting out of sync with whether the tool is/is not in offline mode?

vermeulendivan:
@azvoleff Its working fine on my side (did another check now). But it tended to sometimes work and sometimes not prior to additional changes. So it will be good to do additional testing on your side.

Issue #717, opened 2022-11-10
Enable changing CRS from default of WGS84 for local/remote analyses 


MLNoon:
This one is perhaps more for the developers, but why are custom data imported to trends.earth to calculate sdg 15.3.1 always reprojected to WGS84 lat/lon? There must be a good reason for it, but it is not mentioned in the docs, or I have missed something. Perhaps it is linked to the idea that input data should be processed in GEE? But what if one wants to do it locally? 

Reprojecting GIS data can potentially cause unnecessarily an additional source of uncertainty to the results. Especially it is so when reprojecting from a projected CRS in one datum (native CRS of the source data) to a geographic CRS in another datum. This can cause very annoying results especially when reprojecting sets of different input rasters: inconsistent pixel sizes, inconsistent upper and lower x-y coordinates, inconsistent pixel alignment. Calculations of areas, such as amount of degraded land in hecatres, do not make any sense in geographic CRS such as EPSG:4326.

Could there be in the future possibility to skip the reprojection, and do all calculations in the CRS of user's choosing? Then the tool would be useful also for those, who want to do all calculations locally.

azvoleff:
Adding support here would require:

- adding an option in the plugin settings to choose an output coordinate reference system
- sending the CRS as a WKT string together with the other items in the `params` item of the dictionary of parameters sent for remote tasks
- verifying that user-chosen CRS is used in:
  - **remotely run algorithms**: verifying that the WKT string is used in the GEE code for the remote tasks to reproject results from WGS84 to the user-specified projection (@azvoleff to handle this)
  - **locally run tasks**: reviewing code in [conservationinternational/trends.earth-algorithms](https://github.com/conservationinternational/trends.earth-algorithms) uses the chosen CRS
  - **data imports**: applying chosen CRS when importing data into Trends.Earth

PR #758, opened 2023-01-04
Changing Dataset CRS

gkahiu:
Hi @azvoleff,

Please see the work related to #717. 

For remote tasks, there is a `crs` key added in params that contains the CRS in WKT format using the legacy `WKT1_GDAL` flag. See more options [here](https://api.qgis.org/api/3.28/classQgsCoordinateReferenceSystem.html#ac2bb97dcdc388cfe88479eb836a81b7c) which are dependent on the version of the Proj library.

For locally run tasks, the system now validates whether the input layers are in the same projection as that defined in settings. Also, the `AOI` object from `areaofinterest.prepare_area_of_interest` now uses the CRS from settings. This (AOI) object is cascaded to the respective algorithms.

On supporting importation of datasets, this is not fully implemented as it is dependent on PR #757 but have attempted to refactor some of the sections that explicitly used WGS84. One question is to how to handle the respective input and output resolution computations since there will be no CRS conversion to WGS84 taking place as was the case before.

### DIFF
--- a/LDMP/calculate.py
+++ b/LDMP/calculate.py
@@ -37,6 +37,7 @@ from .conf import AreaSetting
 from .conf import OPTIONS_TITLE
 from .conf import REMOTE_DATASETS
 from .conf import Setting
+from .conf import settings_crs
 from .conf import settings_manager
 from .logger import log
 
@@ -502,7 +503,10 @@ class DlgCalculateBase(QtWidgets.QDialog):
         self.changed_region.emit()
 
     def run_settings(self):
-        self.iface.showOptionsDialog(currentPage=OPTIONS_TITLE)
+        self.iface.showOptionsDialog(
+            self.iface.mainWindow(),
+            currentPage=OPTIONS_TITLE
+        )
         self.update_current_region()
 
     def showEvent(self, event):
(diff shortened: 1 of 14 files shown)


LDMP/data_io.py
```diff
@@ -175,14 +176,19 @@ def __init__(
         self.out_res = out_res
         self.out_data_type = out_data_type
 
+        # TODO: Confirm if this should be optional and if we should have
```
azvoleff, line 179: Yes, we should default to WGS84. Using WGS84 would be consistent with past behavior of the plugin so reasonable to use it as a default.

LDMP/calculate.py
```diff
@@ -522,7 +523,52 @@ def accept(self):
         pass
 
     def settings_btn_clicked(self):
-        self.iface.showOptionsDialog(currentPage=OPTIONS_TITLE)
+        self.iface.showOptionsDialog(self.iface.mainWindow(), currentPage=OPTIONS_TITLE)
+
+    def _validate_crs_multi_layer(self, layer_defn: list) -> bool:
```
azvoleff, line 528: It would be worthwhile to offer users an option to automatically reproject any layers that are in a CRS other than that chosen in their user settings. Otherwise there is no easy way for a user to user layers they may have produced in a different CRS (within Trends.Earth that is), without re-running the analysis. Of course there might be boundary effects, due to the reprojection, there might not be transformations among the CRS if they were user defined, etc., but worth offering it as an option if possible.

LDMP/data_io.py
```diff
@@ -1094,8 +1138,10 @@ def warp_raster(self, out_file):
         log("Importing {} to {}".format(in_file, out_file))
 
         if self.input_widget.groupBox_output_resolution.isChecked():
-            out_res = self.get_out_res_wgs84()
-            resample_mode = self.get_resample_mode(temp_vrt)
+            out_res = None  # self.get_out_res_wgs84()
+            resample_mode = (
+                gdal.GRA_NearestNeighbour
```
azvoleff, line 1143: This changes the logic - in past versions of T.E. `get_resample_mode` was used to use either `gdal.GRA_Mode` for resampling categorical data or `gdal.GRA_Average` for continuous data when aggregating pixels to go from higher to lower resolution. Better to retain that login here rather than always use `gdal.GRA_NearestNeighbour`

azvoleff (commented):
This is looking good overall @gkahiu, I think this is on the right track. One additional comment beyond those inline in the review: changing the CRS in settings breaks the code used to filter layers by their extent for the dropdowns in, for example, the "Indicator for SDG 15.3.1" tool. To address this issue the functions called within `get_usable_datasets` in `data_io.py` will need to be modified. Right now if you run a "Sub-indicators for SDG 15.3.1" job with a different CRS than WGS84, and then try to run "Indicator for SDG 15.3.1" to integrate them, the dropdowns are all empty, even though data should be available and listed in that tool.

PR #767, opened 2023-01-31
updates false positive negative user guide documentation

gdaldegan:


azvoleff:
Thanks @gdaldegan - looks great!

Issue #821, opened 2024-07-04
Internal enhancements and improvements: Add unit tests for core components 


amyburness:
Add unit tests for core components used in the drought and SDG 15.3.1 tools. This will involve adding tests to trends.earth and the trends.earth-algorithms packages. 

- [x] Unit tests added in codebase.

amyburness:
Review with Alex

amyburness:
this ticket has been addressed https://github.com/ConservationInternational/trends.earth/pull/837

PR #837, opened 2024-08-20
Fix CI tests fails

Samweli:
When running unit tests for the remote executed scripts the job requests uses the script id instead of the script name. 
These changes updates the used test data to have script id as the script key when fetching script data.


azvoleff:
The script IDs change when the scripts do (as the IDs are used to ensure the proper script version is run on the API). So to keep the tests up to date in testing the latest version of the scripts we'll need to add a few lines to the `update_script_ids` function within [tasks.py](https://github.com/ConservationInternational/trends.earth/blob/main/tasks.py#L513), to ensure that function also updates the script IDs used in the tests.

@Samweli can you add that update? Once done this one should be good to merge

Samweli:
> The script IDs change when the scripts do (as the IDs are used to ensure the proper script version is run on the API). So to keep the tests up to date in testing the latest version of the scripts we'll need to add a few lines to the `update_script_ids` function within [tasks.py](https://github.com/ConservationInternational/trends.earth/blob/main/tasks.py#L513), to ensure that function also updates the script IDs used in the tests.
> 
> @Samweli can you add that update? Once done this one should be good to merge

sure @azvoleff, thanks.

azvoleff:
Good point - makes sense. I'll merge

amyburness:
#821

PR #849, opened 2024-10-29
Workflow for creating plugin zip artifact on pull requests

Samweli:
Adds new workflow for handling plugin zip file that can be used to check/test/confirm the proposed changes in the PR before merging the PR into the repository.

Screenshot of the comment with a link to the plugin zip file

image


Samweli:
The CI check for the plugin package creation is currently failing because at the moment it is run from this PR context and not from the repository itself. This should be fixed once the PR is merged.

Issue #823, opened 2024-07-04
Integration with third party apps: import functionality for the Land Cover Legend Registry (LCLR) of FAO


amyburness:
Provide an import functionality for the Land Cover Legend Registry (LCLR) of FAO that is based on land cover standard ISO 19144-2, that would provide a link between ISO 19144-series and UNCCD reporting via Trends.Earth.

amyburness:
Deliverables for Integration tasks:
- [ ] Customized import tools for other LPD datasets and for the Land Cover Legend Registry (LCLR) of FAO. 
- [ ] Adapted GUI for FAO-WOCAT. 
- [ ] Documentation of JSON format used to transfer data between Trends.Earth and PRAIS.

mvmaltitz:
As per email from Gabriel on 10 Jan 2025
As for land cover classes that have more than 20 characters, would it be possible to have a tooltip/infotip that shows more characters (not sure we'll be able to show entire names for those classes that have a very long name) when user hover the cursor over classes names?

mvmaltitz:
@gdaldegan 
On the land cover classes names with more than 20 characters, yes, we can show full class names inside tooltips of the corresponding classes that have long names. Should we add tooltips in both the import dialog and the final custom land cover classes table? We could also add another field called full name inside the land cover class information widget that shows the full long class name.

gdaldegan:
@Samweli sounds good, let's do tooltips both in the import dialog and the final custom land cover table. Although, I'm having a hard time to visualize the last suggestion, on adding another field called full name. Let's discuss this in our next sprint (Jan 15th)

mvmaltitz:
## Matrix
Shorter version in the heading and have a tool tip over the name showing the full name

![Image](https://github.com/user-attachments/assets/75d29169-002d-417a-920a-1216d72be70f)

@Samweli to make the update to all the matrix tables

PR #854, opened 2024-11-25
Support for class names import 

Samweli:
Fixes https://github.com/ConservationInternational/trends.earth/issues/823

The new `Import` button under the Custom Land classes group box provides an option to select a `CSV` file and then opens a dialog that allows assignment of parents into the new class names.

The new import button in the Land Cover tool buttons
Screenshot from 2024-11-25 23-25-09

Screenshot of a CSV  file and the resulting import dialog look.

|File|Dialog|
|-----------------------|--------------------------|
|image | image|

### DIFF
--- a/LDMP/conf.py
+++ b/LDMP/conf.py
@@ -59,6 +59,7 @@ class Setting(enum.Enum):
     BASE_DIR = "advanced/base_data_directory"
     CUSTOM_CRS_ENABLED = "region_of_interest/custom_crs_enabled"
     CUSTOM_CRS = "region_of_interest/custom_crs"
+    CSV_FILE_DIR = "advanced/csv_file_dir"
     POLL_REMOTE = "advanced/poll_remote_server"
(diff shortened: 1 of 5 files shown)


amyburness:
@Samweli Feedback from client

- Importing a LCLR example went quite smoothly.
- Import button is a little smaller than the other buttons, it would be good to make it the same size.
- LC class names are not adjusted to show full name once the dialog is expanded (see below)  

image

Samweli:
> * LC class names are not adjusted to show full name once the dialog is expanded (see below)

Thanks @amyburness . The reason the LC class names are not adjusted to show full name is there is a  20 character limit for class names keys, At the moment I have added logic to trim down the LC class names for those who have more than 20 characters. If this is not adequate we can discuss and handle the situation differently.

Samweli:
@gabrieldaldegan any feedback on this PR, I have already updated it to add tooltips to all LC matrix table headers.

gdaldegan:
Thanks @Samweli! Tests ran without issues, it looks good to merge.

Issue #895, opened 2025-04-21
Enable support for adding new periods


dimasciput:
We need to implement functionality that allows users to insert additional progress periods beyond the existing ones.

PR #896, opened 2025-04-21
[WIP] Multiple progress period

dimasciput:
https://github.com/ConservationInternational/trends.earth/issues/895


dimasciput:
[Peek 2025-04-22 16-39.webm](https://github.com/user-attachments/assets/33a1e2f5-7571-4f7b-863d-b66f99e0f98e)

dimasciput:
@gabrieldaldegan @azvoleff

gdaldegan:
thanks @dimasciput, downloaded and tested this implementation. We need to set the time period for the land productivity sub-indicator to use the last 16y of data, as this is a requirement for this sub-indicator specifically to allow to calculate a linear regression and derive the three metrics that compose the land productivity sub-indicator (trend, state, performance). 

For instance, the baseline uses data from 2000 (JRC LPD) or 2001 (Trends.Earth/FAO-WOCAT LDP) to 2015; 1st progress period uses data from 2004 to 2019; and 2nd progress period uses data from 200 to 2023. Please not that this is for Land Productivity only, for Land Cover and SOC the period of interest is the same as the period set across indicators:
image

gdaldegan:
we also need to implement the same solution to "add progress period" under Indicator for SDG 15.3.1, so users can integrate the three sub-indicators for the Baseline, 1st progress period, 2nd progress period and so on:
image

dimasciput:
thanks @gdaldegan for the feedback

> thanks @dimasciput, downloaded and tested this implementation. We need to set the time period for the land productivity sub-indicator to use the last 16y of data, as this is a requirement for this sub-indicator specifically to allow to calculate a linear regression and derive the three metrics that compose the land productivity sub-indicator (trend, state, performance).

Ok, I will update this.

> we also need to implement the same solution to "add progress period" under Indicator for SDG 15.3.1, so users can integrate the three sub-indicators for the Baseline, 1st progress period, 2nd progress period and so on:

Yes, that's on my next to-do list. But I'm trying to build one feature at a time, so this PR only focuses on the sub-indicators. Is that okay? That way, we don't have one big change in a single PR, it's better to keep things smaller.

dimasciput:
> thanks @dimasciput, downloaded and tested this implementation. We need to set the time period for the land productivity sub-indicator to use the last 16y of data, as this is a requirement for this sub-indicator specifically to allow to calculate a linear regression and derive the three metrics that compose the land productivity sub-indicator (trend, state, performance).

@gdaldegan I'm still not quite sure about this. Do you want to add another option to the LPD dataset for 2008–2023?

image

Or are you just looking to enable the productivity dropdown to use the full 16 years of data? Because users should already be able to do that now.

Issue #891, opened 2025-04-10
Documentation of JSON format used to transfer data between Trends.Earth and PRAIS


dimasciput:

azvoleff:
@dimasciput there is now a readthedocs page up on [`te_schemas`](https://github.com/conservationinternational/trends.earth-schemas). That repo contains the schemas used to transfer the tabular data between Trends.Earth and PRAIS. Still building it out but this page gives an overview of how it is used in reporting: https://trendsearth-schemas.readthedocs.io/en/latest/reporting/index.html

dimasciput:
Hi @azvoleff @gdaldegan 

I've been reading the meeting document and listening to the recording, but I'm still not very clear on the goal of the ticket. Is it meant to be just a documentation of the existing JSON summary report, or should I create a new JSON summary file that can be used to transfer data between Trends.Earth and PRAIS?

For example, the new output file could be named trends_earth_prais_report.json, and it would contain data like this (example):

```
{
  "metadata": {
    "title": "Country XYZ - UNCCD SO1 Multi-Year Report",
    "trends_earth_version": "1.6.0",
    "generated_date": "2025-04-18T12:00:00Z",
    "boundary_geojson": "https://example.com/stac/aoi_country_xyz.geojson"
  },
  "land_condition": {
    "Baseline_2000": {
      "year": 2000,
      "land_cover_summary": {
        "Tree-covered areas": 16000,
        "Grasslands": 10500,
        "Croplands": 11000,
        "Wetlands": 2200,
        "Artificial surfaces": 2500,
        "Other Lands": 5200,
        "Water bodies": 1600
      },
      "soc_summary": {
        "average_soc_t_ha": 36.0
      },
      "productivity_summary": {
        "declining_area_km2": 4000,
        "stable_area_km2": 26000,
        "increasing_area_km2": 4000
      }
    },
```

dimasciput:
@gdaldegan Thanks! I think I now have a clearer understanding of this task. I'll create a new JSON for the SO1 form and aim to have something ready by this Friday, even if it's not fully finished, so you can start testing.

PR #898, opened 2025-04-23
Generate SO1-1 and AOI data for PRAIS

dimasciput:
https://github.com/ConservationInternational/trends.earth/issues/891


dimasciput:
Hi @gdaldegan ,

Can you test this? It will generate new JSON data for the SO1-1 form and also create a separate GeoJSON file for the area of interest.
I’ve tried to structure the data to make it easy to export to PRAIS, it should be fairly self-explanatory. But I can also add documentation to show which fields belong to which table, if needed.

After you run the indicator for the SDG, a new folder for PRAIS will be created.

image

gdaldegan:
Thanks for developing this solution! The idea is to have the json files not visible in the front end, as they would not make sense to most users. The purpose of updating the json file is due to the complexity of the original file and making it it less complicated by removing lines for table that are no longer necessary (reflecting the documents shared by UNCCD a while ago) thus making it smoother when importing trends.earth outputs into PRAIS. We will have a dedicated time to go over this in our call with Andrei from Eau de Web, developers of PRAIS.

PR #909, opened 2025-05-15
fix: replace / with _ in branch names during ci packaging

### LINKED ISSUE #908: pr_package CI feature wont run with slashes in branch name
Adding a slash to a branch name will make the package CI fail with:

`FileNotFoundError: [Errno 2] No such file or directory: 'LDMP-fix/901_5b40.zip'`

The problem is that the `/` in the branchname in the filename is interpreted as a path not a sole string.

merydian:
`/` will be replaced with `_` in the filenames now. 

Closes #908.

### DIFF
--- a/.github/workflows/pr_package.yml
+++ b/.github/workflows/pr_package.yml
@@ -40,9 +40,12 @@ jobs:
           invoke set-version
 
       - name: Build zipfile
+        shell: bash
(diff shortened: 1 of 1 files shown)


dimasciput:
@azvoleff can we merge this?

PR #912, opened 2025-05-21
feat: use qgstask for local processes

### LINKED ISSUE #864: Update QgsTask usage
Follow up on the usage of QgsTask class in the plugin background tasks. This should include to identify and resolve all issues

merydian:
Switched from `AbstractWorker` to `QgsTask` in *soilorganiccarbon.py* and *landcover.py*. Other local processes are already running in `QgsTask` via the `LocalJobTask` subclass of `QgsTask`

Closes #864.


dimasciput:
https://github.com/user-attachments/assets/a3e7668b-ecd6-4296-a26f-377152f79174

@merydian I'm still getting the same error.

merydian:
The deletion error also occurs on current main with the landcover and unccd report processes (see #913), but is fixed in 1f62de9faa78844b1e731060d1b8d61dca5a2dc1

Closes #913

Issue #862, opened 2025-01-07
Import custom LC dataset not working properly


gdaldegan:
Only a subset of land cover classes is correctly imported when importing custom land cover datasets developed locally. For instance, tried to import a LC layer for the Bahamas originally containing 8 classes, but only 2 are shown after importing this layer into Trends.Earth using the function.

Samweli:
@gdaldegan can you share the Bahamas layer, so I can test and figure out the underlying issue, thanks.

Samweli:
@gdaldegan does the below screenshot show the expected import workflow and output? 

Screenshot
![image](https://github.com/user-attachments/assets/6b064ffa-bf57-4907-8963-d515e38aebf7)

GIF showing the whole process
![lc_data_import](https://github.com/user-attachments/assets/0ce82db8-eae1-4632-bbd1-8a045d4a0335)

gdaldegan:
thanks @Samweli! Yes, this is the expected output after importing a custom land cover layer. What was preventing it to be correctly imported?

Samweli:
> thanks @Samweli! Yes, this is the expected output after importing a custom land cover layer. What was preventing it to be correctly imported?

@gdaldegan I haven't changed anything in the plugin code yet, I just restored the plugin land cover classes to their default state and ran the process. Can you share the screenshot/steps of you getting the error?

Samweli:
@gdaldegan what are the available LC classes in the settings and the land cover classes aggregation definition you are using when importing?
From my side I use the following

| Custom LC classes(from the plugin settings) |  Aggregation of land cover(from the import dialog) | Output
|-------|---------|------|
|![image](https://github.com/user-attachments/assets/a46a9a66-1daa-4d0c-a681-182ae7b28497)    | ![image](https://github.com/user-attachments/assets/1dc98c0d-0cbb-4c75-b402-42bd78691a56) |![image](https://github.com/user-attachments/assets/6128091a-dac0-4f0f-95bc-4fb87e3164a1)|

gdaldegan:
The land cover pixel values are as follows:
Cropland 0
Grassland 1
Bareground 2
Tree-covered 3
Urban 4
Water 5
Wetland 6

gdaldegan:
![image](https://github.com/user-attachments/assets/f42d5cf9-17c1-4f33-83cc-0e9cab09cd2a)

[...]

PR #928, opened 2025-06-26
add warning message in land cover import dialogue

dimasciput:
image

https://github.com/ConservationInternational/trends.earth/issues/862#issuecomment-2970485085


gdaldegan:
Thanks @dimasciput I think we could follow the same pattern as the message note just below 'Use sample...' checkbox.

dimasciput:
image

@gdaldegan updated

PR #1001, opened 2025-09-26
fix error_recode in compute_unccd

dimasciput:


LDMP/jobs/manager.py
```diff
@@ -98,28 +99,48 @@ def _get_extent_tuple_raster(path):
...
+
+    if any((val > 180 or val < -180) for val in [xmin, xmax]) or any(
+        (val > 90 or val < -90) for val in [ymin, ymax]
+    ):
+        if conf.settings_manager.get_value(conf.Setting.DEBUG):
+            log(
+                f"Failed to calculate extent for {path} - appears undefined/out of bounds "
+                f"({xmin}, {ymin}, {xmax}, {ymax})"
+            )
+        return None
```
azvoleff, line 138: This check won't work in all cases - projected coordinate systems will have xmin/xmax/ymin/ymax outside these bounds. This is an existing bug so fine to keep this check in for now so we can close this issue, but we should create a new issue and fix for this down the road with a more resilient check of the values

dimasciput: thanks for the input @azvoleff

Issue #1106, opened 2026-04-14
Updating Trends.Earth for QGIS 4


dimasciput:
The main goal is to make the Trends.Earth plugin fully functional in QGIS 4, while maintaining backward compatibility (where possible) with QGIS 3.

PR #1107, opened 2026-04-14
Add QGIS 4 compatibility

dimasciput:
https://github.com/ConservationInternational/trends.earth/issues/1106


azvoleff:
Thanks @dimasciput. I tested this locally and also seems to work for me on 4+. Can you rebase this against current head of main?

dimasciput:
Hi @azvoleff, thanks. I have rebased it and tested it again. I’m not sure about the broken security test, but the documentation one is failing because the PR is from my repo. It shouldn’t be an issue once it’s merged.

azvoleff:
Thanks. Looks good - will merge. Thanks for flagging the security scan results @dimasciput: that is a false positive as the QGIS plugins repo marks that one as a warning rather than a blocking item (I need to fix the workflow to reflect this). So can merge as is.

PR #1122, opened 2026-06-01
Fix Qt 6 compatibility and charts

### LINKED ISSUE #1125: QGIS v4 Issues
Image

Dear Nune,

Can we please pass the following message to Federico for the issue on PRAIS4 and Trends.Earth? 

Our issues are as follows:

We used Trends.Earth version 2.3.0, released on May 12, 2026. Since the changelog specifically states that this version “Added support for QGIS 4,” we proceeded with QGIS 4 (works much faster!) for the reporting workflow. However, during the actual report generation process, we encountered several issues under QGIS 4, especially with chart generation (QGIS 3.4 same chart issue after plugin update) and later with the PRAIS4 archive upload.

The first issue was related to chart generation. - Solved

Trends.Earth produced the maps and general report outputs, but all chart PNG files were generated as empty white images. In some cases, the warning log did not show any error and stated that the process was successful, although the chart files were blank.

During troubleshooting, we identified a Plotly-related error in the warning log:

Invalid property specified for object of type plotly.graph_objs.layout.YAxis: 'titlefont'

The error referred specifically to:

titlefont_size

To address this, we manually edited the Trends.Earth plugin file:

C:\Users\User\AppData\Roaming\QGIS\QGIS3\profiles\default\python\plugins\LDMP\reports\charts.py

and replaced:

titlefont_size

with:

title_font_size

After this change, the Plotly property error disappeared.

The second issue was that Plotly was available in the QGIS 4 Python environment, but Kaleido was not installed. When we tested Plotly export directly from the QGIS Python Console, the system returned the following error:
Image export using the "kaleido" engine requires the Kaleido package.

We installed Kaleido manually into the QGIS 4 Python environment using:

"C:\Program Files\QGIS 4.0.2\apps\Python312\python.exe" -m pip install --upgrade kaleido

After this installation, direct Plotly PNG export worked correctly.

However, even after installing Kaleido, Trends.Earth continued to produce white chart PNG files. It appeared that the plugin was still exporting charts through the temporary HTML/QGIS layout export method, rather than using direct Plotly/Kaleido image export. To resolve this, we manually modified the save_image function in charts.py so that it uses:

figure.write_image(...)

After this change, the chart PNGs were generated correctly. We also noticed a possible band reference issue in the land cover chart configuration. The generated land cover raster for our Armenia report contains 6 bands. The relevant bands are:

band 5 = Land_cover_2001
band 6 = Land_cover_2015

However, the chart configuration appeared to refer to:

target_band_num = 19

Since our raster has only 6 bands, we changed the target band to 6. After these changes, the chart files were generated correctly.

The second major issue concerns PRAIS4 upload. - Not Solved

After generating the UNCCD reporting archive from the UNCCD Reporting tab in Trends.Earth, we attempted to upload the archive through the PRAIS4 “Trends.Earth Archive Upload” page. The file uploads successfully, and PRAIS4 shows:

Unpacking spatial layers: Complete

However, when trying to save, PRAIS4 shows:

Save failed

There is no detailed error message indicating which file, spatial layer, or validation rule caused the save failure.

We also tried repacking and cleaning the archive structure, including testing simplified versions, but the same issue remained: the archive uploads and unpacks, but fails at the save step.

In parallel, we tried uploading the spatial layers manually through the PRAIS4 Spatial Layers section. However, PRAIS4 indicates that only one-band raster layers can be uploaded. This seems technically understandable, because the Trends.Earth GeoTIFF outputs are multi-band rasters. For example, the land cover raster includes several bands in one file, such as land cover degradation, ESA classes, transitions, initial land cover, and final land cover.

As a result, direct manual upload of the original Trends.Earth GeoTIFF is not accepted by PRAIS4, while the archive upload also fails at the save stage.

At this stage, the situation is as follows:
We can generate the maps and reports in QGIS 4.
The charts only work after manual changes to the Trends.Earth plugin and manual installation of Kaleido.
The UNCCD reporting archive generated by Trends.Earth uploads and unpacks in PRAIS4, but fails to save.
Manual spatial upload is blocked unless the multi-band Trends.Earth rasters are split into single-band files.
We would appreciate your guidance on the correct workflow for uploading Trends.Earth outputs to PRAIS4, especially in relation to QGIS 4 support, chart generation, multi-band raster handling, and the archive save failure in PRAIS4.

There are other minor issues/inconsistencies along the way as well, but those are not major in face of this one.
We need these issues solved/cleared so the team can work effectively.

Best,
Grigor Janoyan
Team Lead REC Caucasus
 
 
-------- End of forwarded message --------
 
 
Dear Nune, 

Thank you for your detailed and very well-documented message. 

The level of troubleshooting and technical analysis you have carried out is truly impressive, and it is very helpful for us and for the Trends.Earth development team.

Let me try to address each of your points below.

1. Chart Generation Issues

The issues you encountered with chart generation are indeed bugs in the latest Trends.Earth version's support for QGIS 4. 

Specifically:

The titlefont_size → title_font_size change reflects a deprecation in newer versions of the Plotly library. Your fix is correct.
The missing Kaleido package in the QGIS 4 Python environment is an oversight in the plugin's dependency management for QGIS 4.
The save_image function in Trends.Earth was originally designed to use Qt's internal image writer (QImageWriter), which appears to no longer work correctly under the Qt6 framework used by QGIS 4. Your modification to use Plotly's native figure.write_image() is the right approach.
While all three fixes are technically sound, please be aware that they will be overwritten whenever the Trends.Earth plugin is updated. We strongly recommend reporting these issues on the Trends.Earth GitHub repository (https://github.com/ConservationInternational/trends.earth/issues), so that they can be incorporated into a future release.

2. Band Reference Issue (target_band_num = 19)

The target_band_num = 19 value in the chart configuration is likely calibrated for the full multi-band raster output produced by the standard SDG 15.3.1 calculation workflow, which can contain 20 or more bands. If your calculation produced a raster with only 6 bands, this may indicate that a partial or customised calculation was run, or that the output structure differs under QGIS 4. In either case, your adjustment to point to the correct band is a valid workaround, but this discrepancy should also be included in the bug report to the Trends.Earth team.

3. PRAIS4 Archive Upload – "Save Failed"

This is the most critical issue and, unfortunately, it is not something we can resolve on our side. 

The "Save failed" error without further detail is a PRAIS4 platform issue. 

The possible causes include:

The archive structure generated by Trends.Earth may not fully match the schema expected by PRAIS4, particularly if there were changes in the output format with the new version.
PRAIS4 may be performing validation checks on the spatial layers (CRS, resolution, nodata values, file naming conventions) that fail silently.
There may be an incompatibility between the archive produced under QGIS 4 and the expected format.
We recommend the following steps:

a) Open a support ticket on the UNCCD helpdesk at https://support.unccd.int/ — please include the exact archive file you attempted to upload so that the PRAIS4 team can reproduce the issue and provide a detailed error log.

b) As an immediate workaround, we strongly recommend reverting to QGIS 3.x LTR (currently 3.44.10), which is the latest stable and fully documented version. This combination has been tested extensively during the regional workshops and is known to produce archives that PRAIS4 accepts without issues. While QGIS 4 is indeed faster, the QGIS 4 support in Trends.Earth is still very recent and, as your experience demonstrates, not yet fully mature for the complete reporting workflow.

c) If you wish to continue using QGIS 4 for the analysis work (maps, visualisation), you could consider a hybrid approach: use QGIS 4 for exploration and analysis, but switch to QGIS 3 LTR for the final report generation and PRAIS4 archive creation step.

4. Manual Upload of Multi-Band Rasters

You are correct that PRAIS4's manual spatial layer upload only accepts single-band rasters. The Trends.Earth archive upload function is specifically designed to handle the multi-band to single-band splitting automatically during the upload process. Since the archive upload itself is failing, splitting the bands manually using QGIS (Raster → Extraction → Split Bands) and uploading them individually could be attempted as a last resort, but this requires careful attention to file naming conventions and metadata that PRAIS4 expects. Again, we believe the cleanest path is to regenerate the archive using the QGIS 3 LTR. 

If the problem persist using QGIS 3 we should open a ticket to PRAIS to understand exactly what is it causing the issue.

Trying to summarise what are the next steps:

Report the chart generation bugs (titlefont, Kaleido, save_image, band reference) to the Trends.Earth GitHub repository.
Open a PRAIS4 support ticket at https://support.unccd.int/ for the "Save failed" issue, attaching the archive file.
Revert to QGIS 3.44.x LTR  for the reporting archive generation, as this is currently the most reliable path to a successful PRAIS4 upload.
We will also flag these issues with the Trends.Earth development team on our side.
Please do not hesitate to reach out if you need further assistance. 

We are happy to support you throughout the reporting process.

Best regards, 

Federico Gianoli 

JRC Regional Support Hub Joint Research Centre – European Commission

dimasciput:


LDMP/reports/qt_chart_renderer.py
```diff
@@ -0,0 +1,472 @@
...
+import html
+import math
+import re
+import typing
+
+from qgis.PyQt.QtCore import QPointF, QRectF, Qt
+from qgis.PyQt.QtGui import QColor, QFont, QImage, QPainter, QPen
+
+
+class QtChartRenderer:
```
dimasciput, line 12: @azvoleff this to fix the blank charts without having to install Kaleido. 

Image

